PR #216, opened 2014-12-17
Perl Plugin: changed exec in provider 'false' case:

michakrause:
perl -M${module} -e 1 does not work here because of two Problems:

1. $include_dirs is ignored
2. Collectd.pm boundled with collectd uses constants, which are set by the special perl interpreter
   build into collectds perl plugin.
   Without these constants collectd perl modules will not work.

This commit fixes both issues.


michakrause:
fixed

blkperl:
Would it possible to submit some rspec tests? The perl plugin is getting rather complicated and I want to make sure future PRs don't break it.

michakrause:
Sorry, no idea how to do this (yet).

PR #221, opened 2014-12-29
added lvm plugin

sijis:
This add the lvm plugin support.


sijis:
@blkperl just noticed your reply. I've updated the README accordingly.

blkperl:
Thanks @sijis.

PR #222, opened 2015-01-07
arrayify the values and collects

kitchen:
otherwise if you don't pass in an array you get a really awful error :)


nibalizer:
Thanks!

PR #224, opened 2015-01-13
split out the snmp data and hosts into defines

kitchen:
this will make it a lot easier to have a "centralized" collectd snmp
collector that will gather information from network devices and such
which are puppetized but don't run their own collectd :)

### DIFF
--- a/manifests/plugin/snmp.pp
+++ b/manifests/plugin/snmp.pp
@@ -5,8 +5,6 @@ class collectd::plugin::snmp (
   $hosts    = undef,
   $interval = undef,
 ) {
-  validate_hash($data, $hosts)
-
   if $::osfamily == 'Redhat' {
     package { 'collectd-snmp':
       ensure => $ensure,
--- a/manifests/plugin/snmp.pp
+++ b/manifests/plugin/snmp.pp
@@ -1,8 +1,8 @@
 # https://collectd.org/wiki/index.php/Plugin:SNMP
 class collectd::plugin::snmp (
   $ensure   = present,
-  $data     = undef,
-  $hosts    = undef,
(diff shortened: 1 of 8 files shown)


kitchen:
my only thoughts re: changing this would be to move the loading of snmp module out into a lower numbered file, and then put the data configs also < 10, that way you can mix and match between using the params and the defines. Sadly, since the params use capital letters, I can't just wrap those up with create_resources (I tried, sadly did not work)

It would be something like this: 04-snmp-load.conf, 06-snmp-data-<foo>.conf 10-snmp.conf 20-snmp-host-<foo>.conf

txaj:
If I'm not mistaken here we now have both config files generated via parameters (then iterated through in the snmp.conf.erb template) but also defined parameters. 

I invite you to look how I addressed this pattern and avoided that duplication in #168 for postgresql or  #165 for networking
It also "addresses" the order issue

kitchen:
sadly due to the fact that the hash values are [currently](https://github.com/pdxcat/puppet-module-collectd/blob/master/templates/plugin/snmp.conf.erb#L5-L8) [capitalized](https://github.com/pdxcat/puppet-module-collectd/blob/master/templates/plugin/snmp.conf.erb#L13-L17) I can't use `create_resources` in the class to use the defines in the background without breaking backward compatibility :(

The [postgres template](https://github.com/pdxcat/puppet-module-collectd/blob/165815e6d00725ea9b4d6a42d77473529e149b90/templates/postgresql.conf.erb#L7-L34) you changed didn't have that issue, so it was able to be wrapped with `create_resources`

blkperl:
Thanks for the contribution!

PR #226, opened 2015-01-20
java and genericjmx plugins

kitchen:
There's a new global param, which needs to be updated for the other platforms. I only have centos, fedora, debian, ubuntu, and freebsd available to me (main platform is ubuntu, others would be coming from DO boxes) and as far as I can tell genericjmx isn't available for redhat, and java isn't available at all for freebsd, so I'm not overly certain how that should be handled, if there should be a check in the class to see if it's a supported platform or...

Also, I'd love it if someone who is using the collectd java plugin and knows how it behaves when it comes to multiple `<Plugin "java">LoadPlugin "foo" ...` blocks, like if the JVMArg parameter is global across all java plugins, or just that block, or what. The only one I have a need for at the moment is genericjmx, so I tried to make everything generic to support that but there may be some undocumented behavior it does that I'm not aware of since I'm only using one java plugin.

Also, on ubuntu (12.04 anyways) this needs openjdk-6-headless to be installed because it needs that specific jvm's libjvm.so (because they don't have an alternative for libjvm.so, so the package is built against a specific jvm... silly :() so this isn't perfectly drop-in, but I tried :)

I'm all for making changes to this to get this to a mergable state. I am currently using it myself on my ubuntu 12.04 boxes, so it does in fact work, but there's probably a fair amount more work required to get it mergable :)

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -20,7 +20,7 @@ class collectd::params {
       $service_name      = 'collectd'
       $config_file       = "${collectd_dir}/collectd.conf"
       $root_group        = 'root'
-      # FIXME: $java_dir
+      $java_dir          = undef
     }
     'Redhat': {
       $package           = 'collectd'
@@ -40,7 +40,7 @@ class collectd::params {
       $service_name      = 'collectd'
       $config_file       = '/etc/collectd.conf'
       $root_group        = 'root'
-      # FIXME: $java_dir
+      $java_dir          = undef
     }
     'FreeBSD': {
       $package           = 'collectd5'
@@ -50,7 +50,7 @@ class collectd::params {
       $service_name      = 'collectd'
       $config_file       = '/usr/local/etc/collectd.conf'
       $root_group        = 'wheel'
-      # FIXME: $java_dir
+      $java_dir          = undef
     }
     'Archlinux': {
       $package           = 'collectd'
@@ -60,7 +60,7 @@ class collectd::params {
       $service_name      = 'collectd'
       $config_file       = '/etc/collectd.conf'
       $root_group        = 'wheel'
-      # FIXME: $java_dir
+      $java_dir          = undef
     }
     'Gentoo': {
       $package           = 'app-admin/collectd'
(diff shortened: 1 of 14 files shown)


kitchen:
weird, travis failed. I'll look at that in a bit.

manifests/params.pp
```diff
@@ -19,6 +20,7 @@
       $service_name      = 'collectd'
       $config_file       = "${collectd_dir}/collectd.conf"
       $root_group        = 'root'
+      # FIXME: $java_dir
```
blkperl: you can set it to undef for now

manifests/params.pp
```diff
@@ -37,6 +40,7 @@
       $service_name      = 'collectd'
       $config_file       = '/etc/collectd.conf'
       $root_group        = 'root'
+      # FIXME: $java_dir
```
blkperl: set to undef

manifests/params.pp
```diff
@@ -46,6 +50,7 @@
       $service_name      = 'collectd'
       $config_file       = '/usr/local/etc/collectd.conf'
       $root_group        = 'wheel'
+      # FIXME: $java_dir
```
blkperl: same as above

blkperl:
Looks good, just need to fix the failing tests

txaj:
I'm not using that plugin but form what I understand from the doc ( https://collectd.org/documentation/manpages/collectd.conf.5.shtml#plugin_java & https://collectd.org/documentation/manpages/collectd-java.5.shtml ) LoadPlugin instantiate the class with the above JVMArg of the block we're talking about. I believe there is a scoping for each <Plugin java> block

kitchen:
If that's the case that's perfect. Maybe I'll turn collect::plugin::java into a define instead of an almost useless class?

-Jeremy

--
Insert PGP signature here.

kitchen:
sorry I have been busy today and not able to work on this. busybusy! :(

kitchen:
weird, the test failures look like they're coming from concat?

```
     [31m  Undefined variable "::id"; Undefined variable "id" at /home/travis/build/pdxcat/puppet-module-collectd/spec/fixtures/modules/concat/manifests/setup.pp:25 on node testing-worker-linux-docker-2d5c3702-3209-linux-5.prod.travis-ci.org[0m

```

kitchen:
oh, I see the problem. missing facts. fixing.

blkperl:
Please update the README.md with docs for the new plugins and then this is ready to merge

blkperl:
Awesome!

PR #228, opened 2015-01-20
plugin/ntpd: IncludeUnitID option only available in collectd > 5.2

x-way:
Fixes `Plugin 'ntpd' did not register for value 'IncludeUnitID'.` warnings on Debian.


txaj:
:+1:

blkperl:
Thanks for the contribution!

PR #241, opened 2015-02-19
allow to disable forwarding in network::server

devfaz:


blkperl:
Thanks for the contribution!

PR #243, opened 2015-02-25
Package collectd-python is no longer available in EPEL

arioch:
```
# yum info collectd-python
Loaded plugins: fastestmirror
Loading mirror speeds from cached hostfile
 * base: ...
 * epel: ...
 * extras: ...
 * updates: ...
Error: No matching Packages to list
```

Feel free to double check:

https://dl.fedoraproject.org/pub/epel/6/x86_64/
https://dl.fedoraproject.org/pub/epel/7/x86_64/c/

### DIFF
--- a/manifests/plugin/python.pp
+++ b/manifests/plugin/python.pp
@@ -11,12 +11,6 @@ define collectd::plugin::python (
 
   validate_hash($config)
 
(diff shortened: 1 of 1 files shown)


nibalizer:
@blkperl how could they do this to us?

PR #244, opened 2015-02-25
Ruby 1.8.7 is deprecated

arioch:
This patch allows the module to be tested on Ruby 1.8.7 but it won't break the build.

### DIFF
--- a/manifests/plugin/python.pp
+++ b/manifests/plugin/python.pp
@@ -11,12 +11,6 @@ define collectd::plugin::python (
 
   validate_hash($config)
 
-  if $::osfamily == 'Redhat' {
-    package { 'collectd-python':
-      ensure => $ensure,
-    }
-  }
-
   $conf_dir = $collectd::params::plugin_conf_dir
 
   # This is deprecated file naming ensuring old style file removed, and should be removed in next major relese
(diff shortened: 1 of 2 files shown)


manifests/plugin/python.pp
```diff
@@ -11,12 +11,6 @@
 
   validate_hash($config)
 
-  if $::osfamily == 'Redhat' {
-    package { 'collectd-python':
-      ensure => $ensure,
-    }
-  }
```
blkperl: Removed from EPEL7? If so then add some logic so it still works on EPEL6 and EPEL5

blkperl:
I think we should drop ruby 1.8.7 if we don't plan to support it, not just mask the errors.

manifests/plugin/python.pp
arioch: It's removed from 5 and 6 as well.

Source:
https://dl.fedoraproject.org/pub/epel/5/x86_64/
https://dl.fedoraproject.org/pub/epel/6/x86_64/
https://dl.fedoraproject.org/pub/epel/7/x86_64/

arioch:
There you go.

blkperl:
Thanks for the contribution!

PR #246, opened 2015-03-17
Allow non-root users to read plugin configuration

daoloth:
as it uses high ports only, collectd can be run as a non-root user.
moreover, collectd.conf is already 0644.
so why making things different for plugins ?

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -30,7 +30,7 @@ class collectd(
   file { 'collectd.d':
     ensure  => directory,
     path    => $collectd::params::plugin_conf_dir,
-    mode    => '0750',
+    mode    => '0755',
     owner   => 'root',
     group   => $collectd::params::root_group,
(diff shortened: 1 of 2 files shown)


txaj:
No. There could be passwords all over those files. Basic password in the curl plugin, postgresql passwords in the psotgres plugin, etc. 

You can already change the owner of those files if you want to run collectd as another user as root.

daoloth:
"You can already change the owner of those files"

how ?

txaj:
Sorry I oversold this one ; you cannot do it with the code in place.

daoloth:
I know. Are you interested in a pull request that adds the possibility to change the owner of the configuration files (with root as default) ?

txaj:
@blkperl @nibalizer what do you think ? I personally would not make use of it, I use too much plugin requiring root

PR #248, opened 2015-04-04
Fix Rakefile to enforce failure on warnings

blkperl:


txaj:
is it spring already :)

Issue #239, opened 2015-02-15
RHEL7 does not need the collectd-python package


tarjei:
It is included by default.

Thanks for a great module btw!

tarjei:
```diff
--- a/modules/collectd/manifests/plugin/python.pp
+++ b/modules/collectd/manifests/plugin/python.pp
@@ -12,9 +12,9 @@ define collectd::plugin::python (
   validate_hash($config)
 
   if $::osfamily == 'Redhat' {
-    package { 'collectd-python':
-      ensure => $ensure,
-    }
+#    package { 'collectd-python':
+#      ensure => $ensure,
+#    }
   }
 
```

nibalizer:
Can you format this as a pull request?

jesusaurus:
Also, can you remove the package resource rather than commenting it out?

blkperl:
You need to add a conditional on the version of redhat around the package resource.

tarjei:
This wasn't actually meant as a PR, put as a tip to where to change the code :)

bzed:
Using ensure_packages from stdlib would be very appreciated....

blkperl:
Released in #254

Issue #227, opened 2015-01-20
Cannot configure multiple "Import" modules with Python plugin


simonpasquier:
I have a need to instantiate several Python modules and IIUC this isn't supported by the current Puppet manifest.

Ideally it should be possible to generate this configuration:
````
 <Plugin python>
    ModulePath "/path/to/your/python/modules"
    Import "foo"
    Import "bar"

    <Module foo>
      some_param 1
    </Module>
    <Module bar>
      other_param 1
    </Module>
  </Plugin>
````

txaj:
Feel free to submit a PR. The design pattern to use is similar to some other plugins, like postgresql filecount or curl

simonpasquier:
Yep, I'll try to find some time to do it.

blkperl:
Fixed in v3.4 I think

PR #249, opened 2015-04-09
Handle more than one Python module.

bzed:
Github issue #227.

Installing module directories is not handled yet,
only single file modules.

Also fix the RedHat installation of collectd-python for RedHat
Github issue #239.


manifests/plugin/python.pp
```diff
@@ -1,17 +1,19 @@
...
-  $order = '10',
+  $interval = undef,
 ) {
-  include collectd::params
+  require collectd::params
 
-  validate_hash($config)
+  if (
+      $::osfamily == 'Redhat' and
+      operatingsystemmajrelease < 7
```
txaj: does this even gets parsed ?!

manifests/plugin/python.pp
```diff
@@ -20,26 +22,39 @@
...
-      require => File["${name}.load"],
-      notify  => Service['collectd'],
+  $ensure_modulepath = $ensure ? {
+      'absent' => $ensure,
+      default  => 'directory',
+  }
+
+  file { $modulepath :
+      ensure  => $ensure_modulepath,
+      mode    => '0755',
```
txaj: is o+rx necessary ?

txaj:
Could you write some tests as well ? thanks

manifests/plugin/python.pp
bzed: heh, ouch, seems I missed a $::, but yes, it parses well for some reason :)

manifests/plugin/python.pp
bzed: Probably not, but it should not contain sensitive stuff, there are config files for that.

txaj: I'm not a python guy but the least privilege principle should be applied here

bzed:
I'll do if I find the time to do so, unfortunately thats the thing I'm missing most of the time :(

manifests/plugin/python.pp
jesusaurus: Agreed, we shouldn't change the permissions without need.

manifests/plugin/python.pp
```diff
@@ -20,26 +22,39 @@
...
+      owner   => root,
+      group   => root,
```
txaj: should be $collectd::params::root_group

manifests/plugin/python.pp
bzed: the directory is being created. I can't see why there are permissions 'changed'. It is still a folder in /usr/share, there is nothing to hide here.

bzed:
I've fixed the mentioned issues, but the tests still fail - somebody else has to look into it.

mmoll:
try this:
```diff
--- a/spec/defines/collectd_plugin_python_spec.rb
+++ b/spec/defines/collectd_plugin_python_spec.rb
@@ -17,6 +17,7 @@ describe 'collectd::plugin::python', :type => :define do
         :content => "<LoadPlugin \"python\">\n    Globals true\n</LoadPlugin>\n\n\n",
       })
     end
+  end
 
   context ':ensure => absent' do
     let :facts do
```

manifests/plugin/python.pp
```diff
@@ -20,26 +15,39 @@
...
-      mode    => '0640',
-      source  => $script_source,
-      require => File["${name}.load"],
-      notify  => Service['collectd'],
+      require => Package[$collectd::params::package]
+  }
+
+  collectd::plugin { 'python' :
+      ensure   => $ensure,
+      interval => $interval,
```
simonpasquier: I'm pretty sure that the interval setting has no effect on the Python plugin.

txaj: From the LoadPlugin doc : "[interval] Sets a plugin-specific interval for collecting metrics. This overrides the global Interval setting. If a plugin provides own support for specifying an interval, that setting will take precedence."

I am not familiar with Python nor the Python plugin of collectd ; http://collectd.org/documentation/manpages/collectd-python.5.shtml Are the notion of interval in the python plugin mandatory or could they inherit the value from collectd ?

bzed: By default its inherited from collectd, but it is pretty fine to specify a completely different interval. And depending on what you do you in your python module you might want completely different intervals from the default.

simonpasquier: Except that you missed this part: "If a plugin provides own support for specifying an interval, that setting will take precedence.". In the case of the Python plugin, there is no support for custom intervals so the parameter is ineffective. Here is what I get on one of my boxes:

    python plugin: Ignoring unknown config key "Interval"

blkperl:

blkperl:
So #255 was submitted as well. Please work together to figure out which PR should be merged.

blkperl:
Merged #255 which resolves this.

PR #250, opened 2015-04-14
Fix permission on the MySQL plugin file

simonpasquier:
This change modifies the permissions on the MySQL plugin configuration
file to avoid making it readable by everybody.


txaj:
:+1:

Issue #251, opened 2015-04-16
collectd-write_http package doesn't always exist for RedHat derivatives


simonpasquier:
A few months ago, there's been a change (https://github.com/pdxcat/puppet-module-collectd/commit/a334b6e989acb988b2783d90e91c99ab7cff8e88) that installed the collectd-write_http package for RedHat platforms. Unfortunately this package doesn't always exist depending on the repository. For instance, Fedora and EPEL repos have no such package [1] since the write_http plugin is included in the core collectd package.

I suggest to revert the commit (https://github.com/pdxcat/puppet-module-collectd/commit/a334b6e989acb988b2783d90e91c99ab7cff8e88) since it is always possible to install the package outside of the collectd manifest.

[1] https://apps.fedoraproject.org/packages/collectd/overview/

sijis:
I submitted the PR in question here. Sorry for the adverse effect here.

In situations where there's a conflict, I could update those changes to include a param to the class named 'manage_package' and set it to true, as default. This would allow the module to load it but also allow for other modules to control the package installation as desired.

simonpasquier:
@sijis no problem. Indeed adding a class param would be neat to support both cases.

PR #252, opened 2015-04-16
Fix collectd::plugin::write_http for RedHat

simonpasquier:
This change reverts commit a334b6e and closes issue #251. The main RPM
repositories (eg, Fedora and EPEL) don't ship the collectd-write_http
package (instead the plugin is included in the core collectd package).
In case the collectd-write_http needs to be installed, it can be done
outside of the collectd manifests.


blkperl:

blkperl:
Is this going to break Centos5 or Centos6?

simonpasquier:
AFAICT it may break for people that are using custom repositories for installing the collectd packages. Like I said in the initial comment, Fedora and its EPEL derivatives [1] don't ship the write_http plugin as a separate package.

[1] https://apps.fedoraproject.org/packages/collectd/overview/

blkperl:
Thanks for the contribution

PR #254, opened 2015-04-23
Release 3.3.0

blkperl:
Backwards-incompatible changes:

* Drop Ruby 1.8.7 support
* Package collectd-python is no longer available in EPEL

Features:

* Allow disable forwarding in network::server

Bugs/Maint:

* Fix permission on MySQL plugin file
* Only set LogSendErrors option if collectd version >= 5.4.
* CreateFiles/CreateFilesAsync expect boolean values


blkperl:
Need to figure out why the rest of the tests are failing

PR #255, opened 2015-04-23
multiple Python modules support

deric:
Based on postgresql module I rewrote Python modules management. Configuration is generated into two files `10-python.conf` and `11-python.conf` (which should be obviously loaded after the first one).

Usage:
```puppet
  collectd::plugin::python::script {'some-python-module':
    ensure   => $ensure,
    config   => {
      'Host' =>  $host,
      'Verbose' => $verbose,
    }
  }
```
which would generate configuration like this:

```
# Generated by Puppet
<Plugin "python">
  ModulePath "/usr/lib/collectd/python"

  Import "some-python-module"
  <Module "some-python-module">
      Host "localhost"
      Verbose "false"
  </Module>
</Plugin>
```
I wanted to write some specs, but most of current tests fails on my machine. Any hint how could I test this module?

Should fix issue #227.


txaj:
duplicates #249 ... sorry... :)

blkperl:

deric:
Sorry, I didn't see the other one. I guess you can take best from both approaches.

deric:
To follow naming in Python modules, I've renamed the class `collectd::plugin::python::script` to `collectd::plugin::python::module`. "module" shouldn't be reserved word in class naming. Is that ok?

Gemfile
```diff
@@ -8,6 +8,7 @@ group :development, :unit_tests do
   gem 'simplecov', :require => false
   gem 'puppet_facts', :require => false
   gem 'json', :require => false
+  gem 'metadata-json-lint', :require => false
```
blkperl: This doesn't seem to be related to this PR, please make a separate one for this.

blkperl:
Please update the README with docs. Should we keep the old class around for a while with a deprecation warning?

Gemfile
deric: I know, I just wanted to pass Travis tests (won't be necessary if you would run linting and validations after specs).

deric:
There's no way how the old version could be used with multiple modules. I've changed `define collectd::plugin::python` to `class`. To be backward compatible we would have to introduce something like `class collectd::plugin::python2` which might be confusing. I'm not sure if we can have a `define` and a `class` with the same name. If you really want backward compatibility I might try to implement it.

There should be one more spec for the new module define.

txaj:
I don't see either how to keep backward-compatibility without getting the doc sketchy about it.

blkperl:
Ok we can just break backwards compat and note it in the README then. @txaj is this ok to merge?

bzed:
@deric would you mind to take my changes from #249 and merge them? If there is something left to merge? I don't have the time to work on it right now. Thanks!

manifests/plugin/python.pp
```diff
@@ -1,45 +1,51 @@
...
-  $module,
-  $script_source,
-  $ensure = present,
-  $config = {},
-  $order = '10',
+class collectd::plugin::python (
+  $modulepath = '/usr/lib/collectd/python',
+  $ensure     = present,
+  $modules    = {},
+  $order      = '10',
```
txaj: Is this parameter useful in a class context ? (the only other case where we have this is for the perl module)

manifests/plugin/python.pp
```diff
@@ -1,45 +1,51 @@
...
-  validate_hash($config)
+  validate_hash($modules)
 
   $conf_dir = $collectd::params::plugin_conf_dir
 
-  # This is deprecated file naming ensuring old style file removed, and should be removed in next major relese
-  file { "${name}.load-deprecated":
-    ensure => absent,
-    path   => "${conf_dir}/${name}.conf",
+  collectd::plugin {'python':
```
txaj: Should we pass $globals ? 
http://collectd.org/documentation/manpages/collectd-python.5.shtml#configuration
https://github.com/pdxcat/puppet-module-collectd/blob/master/manifests/plugin.pp#L6

manifests/plugin/python.pp
```diff
@@ -1,45 +1,51 @@
...
-      ensure  => $ensure,
-      path    => "${conf_dir}/${order}-${name}.conf",
-      owner   => 'root',
-      group   => $collectd::params::root_group,
-      mode    => '0640',
-      content => template('collectd/python.conf.erb'),
-      notify  => Service['collectd'],
+
+  # should be loaded after global plugin configuration
+  $python_conf = "${collectd::params::plugin_conf_dir}/11-python-config.conf"
```
txaj: You don't need to hardcode '11'. all LoadPLugin directives will be in a file starting with a number - this config file should be called python-config.conf and will be loaded afterwards anyway

manifests/plugin/python/module.pp
```diff
@@ -0,0 +1,34 @@
...
+  $ensure        = 'present',
+  $order         = '50',
+  $python_conf   = "${collectd::params::plugin_conf_dir}/11-python-config.conf",
+){
+  include collectd::params
+  include collectd::plugin::python
+
+  validate_hash($config)
+
+  if $script_source {
```
txaj: is there a case where you would want to configure a python plugin without passing a file ?

manifests/plugin/python/module.pp
```diff
@@ -0,0 +1,34 @@
...
+    file { "${module}.script":
+      ensure => $ensure,
+      path   => "${modulepath}/${module}.py",
```
txaj: A sanity check like validate_absolute_path on that would be apreciated

manifests/plugin/python/module.pp
```diff
@@ -0,0 +1,34 @@
+# Single module definition
+define collectd::plugin::python::module (
+  $config,
+  $script_source = undef,
+  $modulepath    = $collectd::plugin::python::modulepath,
+  $module        = $title,
+  $ensure        = 'present',
+  $order         = '50',
+  $python_conf   = "${collectd::params::plugin_conf_dir}/11-python-config.conf",
```
txaj: I am not sure it's useful to have that parameter, why would we want to have different config files ? (and would it even be handled with the header and footer?)

manifests/plugin/python/module.pp
```diff
@@ -0,0 +1,34 @@
+# Single module definition
+define collectd::plugin::python::module (
+  $config,
+  $script_source = undef,
+  $modulepath    = $collectd::plugin::python::modulepath,
+  $module        = $title,
+  $ensure        = 'present',
+  $order         = '50',
```
txaj: is there a point of being able to order python plugins?

manifests/plugin/python/module.pp
bzed: Yeah, having a whole module directory is prfectly fine. See the netapp collectd python plugin for example. Shipping that directory is the thing that needs to be solved.

txaj:
there's no support for any options inside the <Plugin python> block (Encoding, LogTraces, Interactive)

And I don't know what to think about where the .py files are stored. The approach of @bzed seems prettier but I have not verified wether it's better or not ( https://github.com/pdxcat/puppet-module-collectd/pull/249/files#diff-72cdbba755da633a565633bae5731220R9 )

manifests/plugin/python.pp
deric: Yes, that would be good idea. Fixed in eec219b.

manifests/plugin/python.pp
deric: Ok, fixed.

manifests/plugin/python/module.pp
deric: Nope, not really. AFAIK all the Python plugins are loaded together and the order is not important.

manifests/plugin/python/module.pp
deric: Right, that's not really useful. I'll remove it.

manifests/plugin/python/module.pp
deric: The plugin file could be managed outside of Puppet (e.g. part of a binary package).

manifests/plugin/python/module.pp
deric: A path like `puppet:///modules/myorg/elasticsearch_collectd_python.py` won't pass through `validate_absolute_path()`. Is there some other function that allow paths starting with `puppet://`?

deric:
> there's no support for any options inside the Plugin python block (Encoding, LogTraces, Interactive)

added in 9507b84

Location for Python modules in each distribution is initialized based on @bzed PR. 
I've tried to address all your comments, if there's something missing, let me know.

deric:
According to the documentation I've changed default value for `Globals` to `true`. There's one more thing I forgot, `ModulePath` could be defined multiple times. So, I will change `$modulepath` to an array, it that ok?

txaj:
I'm being picky about the options of the modules because I'm tired to have to do PR (or have to review PR) for options of mofules few days or weeks after a module has been introduced. Let's just do the work well at once :) (this indeed might be a little bit useless for interactive... but anyway)

good catch for ModulePath

deric:
No problem, it's good when there's someone to check the code :) 

About naming conventions, it's better to use `log_traces` or `logtraces`? Then `modulepaths` should follow the same convention. Another "problem" is that `$modulepaths` will now always include default OS path which can't be changed, but you can add other paths where to search for plugins. This allows creating new modules without specifying `modulepath` (OS default is used).

Maybe better approach would be to collect all module paths from each module and then define them in config header. However, this might be over-engineering the whole thing. I'm not sure if anyone would benefit from such feature.

txaj:
We use the very same naming as collectd does.

For $modulepath, we could set the default as [$collectd::params::python_dir] and allows the user to specify something else, wether it includes this path or not. and as a bonus, with your current erb loop this would be very clean if the $modulepath is an empty array :) (which can happen if I understand the doc correctly) 
```
class collectd::plugin::python (
$modulepaths = [$collectd::params::python_dir],
...
```
(Is this valid puppet code ? if not then maybe we could set $collectd::params::python_dir as an array)
this simplifies the code and make the behavior less surprising from outside. (should be explained in the doc though. the doc should also explicit that we are enforcing owner modes and such on those paths)

deric:
>```puppet
class collectd::plugin::python (
  $modulepaths = [$collectd::params::python_dir],
```

`$collectd::params::python_dir` is not imported at that line, it would result into `ModulePath "undef"`. I've removed "magic" merging with paths with defaults. Now you can override `$modulepaths`.

**Comment by deric** [comment]
@txaj any progress on this? Is there some problem remaining?

File: manifests/plugin/python.pp [hunk 9]
```diff
@@ -1,45 +1,83 @@
...
+  validate_hash($modules)
+  validate_bool($interactive)
+  validate_bool($logtraces)
+  validate_bool($globals)
+  validate_array($modulepaths)
+
+  $module_dirs = empty($modulepaths) ? {
+    true  => [$collectd::params::python_dir],
+    # use paths provided by the user
+    false => $modulepaths
```
txaj: If we pass an array to `$modulepaths` it blows up
```
    context ':ensure => present and multiple $modulepaths' do
      let :params do
        {
            :modulepaths => ['/tmp/', '/data/']
        }
      end
      it 'will include two directories' do
        should contain_file('/tmp/')
        should contain_file('/data')
      end
    end
```

txaj: my bad I forgot a trailing slash

txaj:
@deric please squash your commits and you're good to go

deric:
@txaj squashed.

txaj:
@blkperl this is good for me

blkperl:
Thanks for the contribution!

txaj:
yeah thanks for the good work @deric

deric:
Your welcome. I've added a note in the [README](https://github.com/pdxcat/puppet-module-collectd/pull/255/files#diff-04c6e90faac2675aa89e2176d2eec7d8R661), that this feature will be part of 3.4 release. Just remember to change it to whichever version it will be part of :)

PR #256, opened 2015-04-23
Fix tests

txaj:
Not sure why but this did the trick :
https://github.com/rodjek/rspec-puppet/issues/205#issuecomment-45907885

### DIFF
--- a/spec/classes/collectd_init_spec.rb
+++ b/spec/classes/collectd_init_spec.rb
@@ -30,7 +30,7 @@ describe 'collectd' do
    end
 
    it 'should fail' do
-     expect { subject }.to  raise_error(/foo is not supported/)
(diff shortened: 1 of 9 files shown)


spec/classes/collectd_init_spec.rb
```diff
@@ -30,7 +30,7 @@
    end
 
    it 'should fail' do
-     expect { subject }.to  raise_error(/foo is not supported/)
+     should compile.and_raise_error(/foo is not supported./)
```
nibalizer: why the . here?

txaj: lousy copy-paste. tests are broken anyway for other reasons in travis... i'm on it right now

Issue #261, opened 2015-05-05
Support multiple graphite backends


deric:
Currently only single `<Carbon>` instance could be specified.

blkperl:
Fixed in v4.0.0

PR #262, opened 2015-05-05
Support multiple carbon backends

deric:
Basic implementation of #261.

Very similar to Python PR.  You can define mupliple carbon backend using `carbon` define, example usage:
```
  collectd::plugin::write_graphite::carbon {'some_unique_graphite_identifier':
    graphitehost   => 'localhost',
    graphiteport   => 2003,
    graphiteprefix => '',
  }
```

Is it OK to drop old interface? All parameter names stayed the same. Just opening for discussion, there's still a lot of work on the specs.


txaj:
You have a real use case for that ? Breaking the interface is ok in that case, but if you see more plugin to modify in that direction then it'd be better to modify them at the same time. Breaking interfaces like this calls for a major release each time

deric:
Yes, there are multiple databases that supports graphite protocol (e.g. InfluxDB). It could be used for failover servers or when you try to optimize Graphite backend storages (there's Whisper, Ceres, etc.) and you don't want to loose your metrics.

The documentation says:
> The configuration consists of one or more `<Node Name>` blocks

Right now the block is called `<Carbon>` I'll rename that to avoid problems with future versions.

txaj:
Very good point :) 
Don't forget to handle the old collectd versions. This changes seems to have happened between [5.2](https://github.com/collectd/collectd/blob/collectd-5.2/src/collectd.conf.pod#plugin-write_graphite) and [5.3](https://github.com/collectd/collectd/blob/collectd-5.3/src/collectd.conf.pod#plugin-write_graphite)

bzed:
From my experience I'd suggest not to let collectd handle such things, but install carbon-c-relay on the same machine and configure it accordingly. It will work fine in collectd, though.

deric:
@bzed I understand you objections, some might object that Graphite is very good storage. Anyway, you don't have to use this feature :) This PR simply implements Collectd API. 

For Collectd >= 5.3 `<Node "title">` syntax is used. In case that `@collectd_version` is not defined, which one should be used?

txaj:
The most common implementations I've seen in this module write the legacy way when ```@collectd_version``` is not known yet

deric:
@txaj Ok, fixed.

blkperl:
Woot! Thanks for the contribution

PR #265, opened 2015-05-06
allow for list of timerpercentiles

nathanielc:


txaj:
interesting way not to break the interface, however 1/ do we want to do that ? (be that loose with types) 2/ there's no docs

ghoneycutt:
Looks good to me, but needs docs, spec tests and validation of `$timerpercentile`

templates/plugin/statsd.conf.erb
```diff
@@ -17,7 +17,11 @@
 <% unless @deletesets.nil? -%>
   Deletesets <%= @deletesets %>
 <% end -%>
-<% if @timerpercentile -%>
+<% if @timerpercentile.kind_of?(Array) -%>
+<% @timerpercentile.each do |p| -%>
+  TimerPercentile <%= p %>
+<% end -%>
+<% elsif @timerpercentile -%>
```
igalic: why not just,
```ruby
<% [@timerpercentile].flatten.compact.each do |p| -%>
  TimerPercentile <%= p %>
<% end -%>
```

blkperl:
Closing due to lack of activity. Please reopen when the feedback has been addressed

PR #270, opened 2015-06-09
add new cpu plugin options introduced in collectd 5.5

piotr1212:
This makes new cpu options configurable, see https://collectd.org/documentation/manpages/collectd.conf.5.shtml#plugin_cpu

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,7 +157,9 @@ class { 'collectd::plugin::conntrack':
 ```
 
 ####Class: `collectd::plugin::cpu`
-
+  reportbystate => true,
+  reportbycpu => true,
+  valuespercentage => true,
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -157,7 +157,9 @@ class { 'collectd::plugin::conntrack':
 ```
 
 ####Class: `collectd::plugin::cpu`
-
+  reportbystate => true,
+  reportbycpu => true,
+  valuespercentage => true,
```
blkperl: You should note it only works in versions greater than 5.5

blkperl:
Can you add a quick spec test as well?

blkperl:
Thanks for the contribution!

PR #278, opened 2015-06-19
Use concat to define the process plugin config.

bzed:
Adding a parser function to keep it compatible with the old way
to define process matches.


bzed:
Somebody please fix the spec tests, I'm blind and don't spot the error.

manifests/plugin/processes.pp
```diff
@@ -10,7 +11,43 @@
 
   collectd::plugin {'processes':
     ensure   => $ensure,
-    content  => template('collectd/plugin/processes.conf.erb'),
+    order    => $order,
     interval => $interval,
   }
+
+  concat{"${collectd::params::plugin_conf_dir}/processes-config.conf":
```
txaj: I think you're missing a ```include collectd::params```

bzed: Good catch!

blkperl:
I'm slightly confused on what this patch does? Do the docs need to be updated as well?

bzed:
I've updated the documentaion, hope its clear now?

blkperl:
Ok it makes sense now. Thanks for the contribution

PR #279, opened 2015-06-19
add support for protocols plugin

throck:
Adding support for the protocols plugin might be useful for others


blkperl:
Thanks for the contribution!

PR #280, opened 2015-06-24
fix cpu config parsing

bzed:
The current cpu plugin breaks older collectd versions as they fail to parse the <Plugin> ...</Plugin> stuff.


txaj:
:+1:

blkperl:
Thanks for the contribution!

Issue #281, opened 2015-06-29
Wrong format in python plugin configuration


ttarczynski:
## Issue description

After upgrading to pdxcat/collectd 3.4.0 I get wrong formatting in the python plugin configuration.

* puppet code:

```
    collectd::plugin::python::module {'rabbitmq':
      modulepath    => '/usr/lib64/collectd',
      script_source => 'file:///usr/lib64/collectd/rabbitmq.py',
      config        => {
        'Username'    => 'uuu',
        'Password'    => 'ppp',
        'Realm'       => 'RabbitMQ Management',
        'Host'        => 'localhost',
        'Port'        => '15672',
        'PrefixVhost' => true,
      }
    }
```

* generated configuration file:

```
# cat /etc/collectd.d/python-config.conf 
# Generated by Puppet
<Plugin "python">
  ModulePath "/usr/lib64/collectd"
  LogTraces false
  Interactive false

  Import "rabbitmq"

  <Module "rabbitmq">
      Host localhost      Password ppp      Port 15672      PrefixVhost true      Realm RabbitMQ Management      Username uuu  
  </Module>

</Plugin>
```

* I think the `<Module "rabbitmq">` block should be formatted like this:

```
  <Module "rabbitmq">
      Host "localhost"    
      Password "ppp"
      Port "15672"    
      PrefixVhost true
      Realm "RabbitMQ Management"
      Username "uuu"
  </Module>
```
## Environment

```
# puppet --version
3.7.4
# ruby -v
ruby 1.8.7 (2013-06-27 patchlevel 374) [x86_64-linux]
# cat /etc/redhat-release 
CentOS release 6.6 (Final)
```

## Suggested cahnges

* add new line after each param
* add double quotes for values that are strings (but not for numbers and booleans)

---

I will try to submit a PR with this change.

ttarczynski:
I've submitted PR #283 to add missing newlines.
And I've learned I can get double quotes with the following puppet code:

```
    collectd::plugin::python::module {'rabbitmq':
      modulepath    => '/usr/lib64/collectd',
      script_source => 'file:///usr/lib64/collectd/rabbitmq.py',
      config        => {
        'Username'    => '"uuu"',
        'Password'    => '"ppp"',
        'Realm'       => '"RabbitMQ Management"',
        'Host'        => '"localhost"',
        'Port'        => '"15672"',
        'PrefixVhost' => 'true',
      }
    }
```

blkperl:
[...]

PR #282, opened 2015-06-29
Fix formatting in python plugin template

ttarczynski:
Fix for: https://github.com/pdxcat/puppet-module-collectd/issues/281

### DIFF
--- a/templates/plugin/python/module.conf.erb
+++ b/templates/plugin/python/module.conf.erb
@@ -1,7 +1,11 @@
   Import "<%= @module %>"
 
   <Module "<%= @module %>">
-  <% @config.sort.each do |key,value| -%>
-    <%= key -%> <%= value -%>
-  <% end %>
+  <%- @config.sort.each do |key,value| -%>
+  <%- if value.is_a?(String) -%>
+    <%= key %> "<%= value %>"
+  <%- else -%>
+    <%= key %> <%= value %>
+  <%- end -%>
+  <%- end -%>
   </Module>
--- a/templates/plugin/python/module.conf.erb
+++ b/templates/plugin/python/module.conf.erb
@@ -2,10 +2,6 @@
 
   <Module "<%= @module %>">
   <%- @config.sort.each do |key,value| -%>
-  <%- if value.is_a?(String) -%>
-    <%= key %> "<%= value %>"
-  <%- else -%>
     <%= key %> <%= value %>
-  <%- end -%>
   <%- end -%>
   </Module>


ttarczynski:
The failed tests show that the python module should be used with the double quotes included in parameters values:

```
  context 'spam module' do
    let(:title) { 'spam' }
    let :params do
      { 
        :config     => { 'spam' => '"wonderful" "lovely"' },
        :modulepath => '/var/lib/collectd/python',
      }
    end
```

```
      should contain_concat__fragment('collectd_plugin_python_conf_spam').with({
        :content => /spam "wonderful" "lovely"/,
      })
```

---

I think this behavior has changed since version 3.3.0 and it would help if we have a note in the documentation.
Should I revert the double quotes change? Then I would only add new lines after each param and a short note in docs about the quotes.

blkperl:
@deric Did your recent PR break this?

ttarczynski:
I think collectd configuration syntax hasn't changed.
I've meant a change in puppet-module-collectd between version 3.3.0 and 3.4.0

The rspec code ([collectd_plugin_python_module_spec.rb](https://github.com/pdxcat/puppet-module-collectd/blob/master/spec/defines/collectd_plugin_python_module_spec.rb)) shows configuration similar to the man page example: [collectd-python](https://collectd.org/documentation/manpages/collectd-python.5.shtml#synopsis). But there is the missing newlines issue (#281)

deric:
Nope, that wasn't me. The line ending was removed in [#267](https://github.com/pdxcat/puppet-module-collectd/commit/11f42ad491b7c16f7118a0ce6c2b017d989cd3b6).

arioch:
#267 fixed not getting any python based data. It's still happily running in production over here.

templates/plugin/python/module.conf.erb
```diff
@@ -1,7 +1,11 @@
   Import "<%= @module %>"
 
   <Module "<%= @module %>">
-  <% @config.sort.each do |key,value| -%>
-    <%= key -%> <%= value -%>
-  <% end %>
+  <%- @config.sort.each do |key,value| -%>
+  <%- if value.is_a?(String) -%>
+    <%= key %> "<%= value %>"
```
deric: Be careful, this doesn't check if string is already quoted. Have a look at the [specs](https://github.com/pdxcat/puppet-module-collectd/blob/master/spec/classes/collectd_plugin_python_spec.rb#L90).

templates/plugin/python/module.conf.erb
```diff
@@ -1,7 +1,11 @@
...
+  <%- else -%>
+    <%= key %> <%= value %>
+  <%- end -%>
+  <%- end -%>
```
deric: Should be replaced by `<% end %>` if you don't wanna re-introduce #267.

deric:
@arioch sure, but removing line endings after key-value pair (`<%= key -%> <%= value -%>`) wasn't necessary.

ttarczynski:
I'm not sure if I'm using the pull request correctly. I've just added another commit, to change only newlines (and not quotes).

@arioch @deric I think the code on master works ok for a single parameter in `Module` block, but it doesn't for multiple params, e.g. #281

templates/plugin/python/module.conf.erb
```diff
@@ -1,7 +1,7 @@
   Import "<%= @module %>"
 
   <Module "<%= @module %>">
-  <% @config.sort.each do |key,value| -%>
-    <%= key -%> <%= value -%>
-  <% end %>
+  <%- @config.sort.each do |key,value| -%>
+    <%= key %> <%= value %>
+  <%- end -%>
```
deric: Please replace this by `<% end %>`, otherwise there won't be new line between multiple modules.

templates/plugin/python/module.conf.erb
ttarczynski: I think #267 was caused by a missing newline after `</Module>`

templates/plugin/python/module.conf.erb
ttarczynski: I will check how it works for multiple modules.
For only one module I get this config file:

```
# Generated by Puppet
<Plugin "python">
  ModulePath "/usr/lib64/collectd"
  LogTraces false
  Interactive false

  Import "rabbitmq"

  <Module "rabbitmq">
    Host "localhost"
    Password "ppp"
    Port "15672"
    PrefixVhost true
    Realm "RabbitMQ Management"
    Username "uuu"
  </Module>

</Plugin>
```

ttarczynski: Just tested with multiple modules. 
It results with config file:

```
# Generated by Puppet
<Plugin "python">
  ModulePath "/usr/lib64/collectd"
  LogTraces false
  Interactive false

  Import "elasticsearch"

  <Module "elasticsearch">
    Cluster test
    Verbose false
    Version 1.0
  </Module>

  Import "rabbitmq"

  <Module "rabbitmq">
    Host "localhost"
    Password "ppp"
    Port "15672"
    PrefixVhost true
    Realm "RabbitMQ Management"
    Username "uuu"
  </Module>

</Plugin>
```

I think newlines between modules looks ok here.

ttarczynski: @deric when I replace `<%- end -%>` by `<% end %>` I get extra (not needed) new line before `</Module>` and wrong indentation in the `<Module>` block:

```
Notice: /Stage[main]/Collectd::Plugin::Python/Concat[/etc/collectd.d/python-config.conf]/File[/etc/collectd.d/python-config.conf]/content: 
--- /etc/collectd.d/python-config.conf  2015-06-30 09:45:40.752012515 +0200
+++ /tmp/puppet-file20150703-10516-16ym6qz-0    2015-07-03 08:50:02.977012923 +0200
@@ -8,19 +8,21 @@
 
   <Module "elasticsearch">
     Cluster opslogstash
-    Verbose false
-    Version 1.0
+      Verbose false
+      Version 1.0
+  
   </Module>
 
   Import "rabbitmq"
 
   <Module "rabbitmq">
     Host "localhost"
-    Password "ppp"
-    Port "15672"
-    PrefixVhost true
-    Realm "RabbitMQ Management"
-    Username "uuu"
+      Password "ppp"
+      Port "15672"
+      PrefixVhost true
+      Realm "RabbitMQ Management"
+      Username "uuu"
+  
   </Module>
 
 </Plugin>
```

deric:
Ok, cool. Thanks for debugging this.

blkperl:
So is this ready to be merged then?

deric:
Yeah, good to merge. I've tested this on a configuration with multiple python modules. Indentation and new lines works fine. Tested on `puppet 3.8.1`

PR #285, opened 2015-07-03
fix multiple instances for curl_json

d3cker:
I need multiple curl_json instances which is generated automaticly. 
Use of ensure_packages() allow to call collectd::plugin::curl_json more than once.


blkperl:
Thanks for the contribution!

Issue #97, opened 2014-02-01
Add rspec puppet tests for plugins that don't have them


blkperl:

jyaworski:
This is way too broad; please open an issue with individual plugins.

PR #287, opened 2015-07-04
support definition of multiple exec commands

deric:
A breaking change because `collectd::plugin::exec` is renamed to `collectd::plugin::exec::cmd` (just to follow conventions in this module). Otherwise it wouldn't be hard to keep compatibility, because `exec` was already implemented as a `define`.


deric:
Added tests for exec plugin (partially solves #97).

Fixed case when empty `exec` array is passed (`Exec` statement shouldn't be generated).

blkperl:
Thanks for the contribution!

PR #289, opened 2015-07-12
add support for aggregation plugin and chains

kurpipio:


blkperl:
Some spec tests would be nice to ensure no one breaks it in the future but I'll merge it for now.

blkperl:
Thanks for the contribution!

PR #291, opened 2015-07-14
use ensure_packages for package[collectd-curl_json] in collectd::plugin::curl_json

zivis:
if this define is multiple used on RedHat systems the package[collectd-curl_json] is declared multiple times and the compilation will fail.  ensure_packages (ensure-resource) will avoid this.


d3cker:
Hello, I've already reported it here: https://github.com/pdxcat/puppet-module-collectd/pull/285

PR #296, opened 2015-07-23
Add carbon default values as parameter to create_resources

jyaworski:
Hello:

This allows default values to be passed in for carbon blocks, and it is useful in the event that you have many unique carbon backends.

Critique welcome.

Inspiration taken from https://github.com/jfryman/puppet-nginx/blob/master/manifests/init.pp#L300


blkperl:
Thanks for the contribution!

PR #297, opened 2015-07-24
adding ValuesAbsolute and ValuesPercentage to memory and swap

ChrisHeerschap:
Memory now has the ability to report in absolute values or percentages. I've tested these changes and they work for me.


blkperl:
Please write a spec tests that verifies the conditional. It looks like your changes might break the old behavior.

ChrisHeerschap:
Re: writing the spec - you mean something along the line of ./spec/classes/collectd_plugin_swap_spec.rb?

I'll see what I can do - still very new to puppet so learning as I go.

ChrisHeerschap:
added the same Values{Absolute,Percentage} to swap, and tested that to work on CentOS 6.6 using collectd 5.5

ChrisHeerschap:
Copying spec file from swap, I see stuff I didn't change. I'll get it fixed.

ChrisHeerschap:
Okay, I've fixed the swap spec to check for version 5.5 adding the ValuesPercentage, and I've created a memory spec. Not sure how I would test these locally so I had to commit after each and hope for the best, but I've got the memory spec working. One thing that I tried to do - unsuccessfully - was to check that a version below 5.5 would not have the ValuesPercentage, but that spec failed. I tested it manually using collectd 4.10.9 from EPEL and it worked correctly and logs memory metrics without complaint.

ChrisHeerschap:
Okay, now I think I have the spec file checking the conditional properly - making sure that the 10-memory.conf doesn't have ValuesPercentage at a version prior to 5.5. All checks passed, give it a look and let me know if I'm missing something. Thanks!

blkperl:
Looks good, thanks for the contribution!

ChrisHeerschap:
Thank you, that was a valuable learning experience.

PR #299, opened 2015-07-27
add collectd::plugin::ceph

TheMeier:
ceph plugin support
create custom fact to find ceph osd instances
create ceph plugin config using the custom fact


lib/facter/collectd_discoverd_osds.rb
```diff
@@ -0,0 +1,18 @@
...
+#
+# Purpose: discover ceph osds
+#
+# Resolution:
+#
+# Caveats:  not well tested
+#
+require 'puppet'
+
+if File.directory?('/var/lib/ceph/osd')
```
txaj: is this multi-platform ?

txaj:
this PR relates to : https://collectd.org/documentation/manpages/collectd.conf.5.shtml#plugin_ceph

I think this is a bit too much implementation dependent... Also this plugin cannot be used for monitors nor mds. 
This is a great piece of puppet code but not modular enough to be accepted in a module like this. Especially I don't think it's this module' role to provide a fact to discover OSDs

In this module should only be PR'ed code to create the configuration from puppet args, with a hash containing daemon names & paths. Your fact and a configuration snippet using it could be given in the doc though

TheMeier:
Maintaining a hash of which osd/mon instance is running where is not really feasible with ceph. Just imaging having a ceph cluster with tens or even hundreds of nodes and 10-50 osds per node. The osd names being randomly distributed and osds being added removed over time. Even if you maintain the osd instances with puppet, you specify them with the disk they use and don't know the resulting osd names. Actually i don't really see an alternative to autodiscovery.

As for cross platform, as far as I know ceph is only available on linux and all the path are quite static for now.

txaj:
You are viewing the problem with problem-solution mindset. I'm viewing it by recalling the "best practice" for keeping maintainable puppet code. 

But you might be right ; the setup you describe is the one where puppet would be extremely useful. 

Could someone ( @blkperl ? ) validate that we'd welcome a plugin that is responsible of both discovering the service it's supposed to monitor *and* to setup that monitoring ? If yes, then @TheMeier we'd like you to PR documentation as well as some basic testing (but I'd prefer you to get some approval before working further in that direction).

lib/facter/collectd_discoverd_osds.rb
daenney: The `File.directory?()` call is because Ruby uses forward slashes for path separators on any OS but I don't think that that's the path for Windows for example, if someone managed to pull that off.

nibalizer:
The fact could realistically live in the puppet-ceph module http://git.openstack.org/cgit/stackforge/puppet-ceph/

I would agree that I prefer puppet code that takes orders and does them. Puppet code that attempts to perform autodiscovery is usually pretty janky. Making this code work such that the array of osd's is passed in to the template/class I think will make everyone happy, those that want it to be autodiscovering can simply use the fact to detect the osds then pass that variable in.

TheMeier:
I changed it to use an provided array, also added some docs and cleanup

blkperl:
Awesome, can you squash the commits and then I will merge it

https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit

TheMeier:
Hm that merge commit was a bad idea. 
I have a clean version now here https://github.com/TheMeier/puppet-module-collectd/tree/ceph_plugin_2 should i create a new PR?

daenney:
Please, yes.

PR #301, opened 2015-07-27
Support UdevNameAttr attribute on disk plugin (fixes #300)

### LINKED ISSUE #300: Support new disk plugin attribute 'UdevNameAttr'
The disk plugin in collectd 5.5 has a new attribute called `UdevNameAttr`, which allows it to look up disk names via udev:

https://collectd.org/documentation/manpages/collectd.conf.5.shtml#plugin_disk

This is really useful on systems with logical volumes and so it would be good to have support for it in `collectd::plugin::disk`.

jonnangle:


blkperl:
Thanks for the contribution!

PR #302, opened 2015-07-29
Allow iptables chains parameter to be an array (fix #237)

### LINKED ISSUE #237: iptables module needs to support multiple chains by same name
I would like to be able to do:

    class { 'collectd::plugin::iptables':
      chains  => {
        'filter' => 'INPUT',
        'filter' => 'OUTPUT',
      },
    }

In my config, but it only writes the OUTPUT(/last) option.

jonnangle:
When using the iptables plugin, this allows you to provide more than one chain for each table by passing an array as the hash value. For example:

```puppet
class { 'collectd::plugin::iptables':
  chains  => {
    'filter' => ['INPUT','OUTPUT'],
  },
}
```

### DIFF
--- a/templates/plugin/iptables.conf.erb
+++ b/templates/plugin/iptables.conf.erb
@@ -3,7 +3,7 @@
 <%   @chains.each_pair do |table,chains|
        Array(chains).each do |chain| -%>
   Chain <%= table %> <%= chain %>
-<%     end
-     end -%>
+<%     end -%>
+<%   end -%>
 </Plugin>
 <% end -%>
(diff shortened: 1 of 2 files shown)


templates/plugin/iptables.conf.erb
```diff
@@ -1,7 +1,9 @@
 <% if @chains -%>
 <Plugin iptables>
-<%   @chains.each_pair do |table,chain| -%>
+<%   @chains.each_pair do |table,chains|
+       Array(chains).each do |chain| -%>
   Chain <%= table %> <%= chain %>
-<%   end -%>
+<%     end
+     end -%>
```
blkperl: Isn't that missing a `<%` why is Travis not failing?

jonnangle: Agreed, now fixed!

blkperl:
Thanks for the contribution!

PR #303, opened 2015-08-03
add option to not install collectd-iptables package

gibre:
Needed on centos 6 for example, where the package does not exist at all.


blkperl:
Thanks for the contribution!

PR #304, opened 2015-08-05
Add support for the AllPortsSummary option added in collectd 5.5.0.

dfxprima:
Collectd version 5.5.0 added the AllPortsSummary option to the tcpconns module.  This patch adds support for that option.

### DIFF
--- a/manifests/plugin/tcpconns.pp
+++ b/manifests/plugin/tcpconns.pp
@@ -1,11 +1,11 @@
 # https://collectd.org/wiki/index.php/Plugin:TCPConns
 class collectd::plugin::tcpconns (
-  $localports  = undef,
-  $remoteports = undef,
-  $listening   = undef,
-  $interval    = undef,
-  $summary     = undef,
-  $ensure      = present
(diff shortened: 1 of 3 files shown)


blkperl:
Can you add a spec test?

dfxprima:
I'll give it a try.  I've never written rspec before so it may take me while.

blkperl:
You can look at the other plugins for examples of how to do the version check

https://github.com/pdxcat/puppet-module-collectd/blob/master/spec/classes/collectd_plugin_write_graphite_spec.rb#L96

dfxprima:
I renamed to the summary variable to allportssummary so it matches the collectd tcpconns parameter name.  I also added the tests to validate the value of allportssummary and its inclusion in the generated file based on collectd version.

blkperl:
Please squash your commits and rebase against master.

dfxprima:
Please close/cancel this request.   I'll start over by putting the changes on a branch like I should have in the first place.

Thanks!

PR #305, opened 2015-08-05
Improve version handling to avoid multiple puppet runs for some situa…

bdellegrazie:
…tions

When an administrator knows the version of collectd that will be used or
at least the minimum version available the need for two puppet runs before
convergence can be avoided or at least minimised.

Instead of using the fact in the templates they now use a class variable
set to one of (in priority order):
* collectd_version (i.e. the fact)
* version (the semver matched part of it only)
* minimum_version (undef by default)

Existing behaviour is preserved except for a corner case where version is
set to something specific and collectd is not yet installed. In this case
puppet will only take one run and assume the version specified when creating
the templates

references pdxcat/puppet-module-collectd#162

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,9 +27,10 @@ declaration:
 
 ```puppet
 class { '::collectd':
-  purge        => true,
-  recurse      => true,
-  purge_config => true,
+  purge           => true,
+  recurse         => true,
+  purge_config    => true,
+  minimum_version => '5.4',
 }
 ```
 
@@ -37,6 +38,10 @@ Set purge, recurse, and purge_config to true in order to override
 the default configurations shipped in collectd.conf and use
 custom configurations stored in conf.d. From here you can set up
 additional plugins as shown below.
+
+Specifying the version or minimum_version of collectd as shown above reduces the need for
+two puppet runs to coverge. See [Puppet needs two runs to correctly write my conf, why?](#puppet-needs-two-runs-to-correctly-write-my-conf,-why?) below.
+
 
 Simple Plugins
 --------------
@@ -1168,7 +1173,12 @@ See metadata.json for supported platforms
 
 ##Known issues
 
-Some plugins will need two runs of Puppet to fully generate the configuration for collectd. See [this issue](https://github.com/puppet-community/puppet-collectd/issues/162).
+###Puppet needs two runs to correctly write my conf, why?
+
+Some plugins will need two runs of Puppet to fully generate the configuration for collectd. See [this issue](https://github.com/pdxcat/puppet-module-collectd/issues/162).
+This can be avoided by specifying an explicit version (`$version`) or a minimum version (`$minimum_version`) for the collectd class. e.g. Setting either of these to 1.2.3 will
+make this module assume on the first run (when the fact responsible to provide the collectd version is not yet available) that your systems are running collectd 1.2.3
+and generate the configuration accordingly.
 
 ##Development
 
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -15,11 +15,21 @@ class collectd(
   $write_queue_limit_low  = undef,
   $package_name           = $collectd::params::package,
   $version                = installed,
+  $minimum_version        = undef,
 ) inherits collectd::params {
 
   $plugin_conf_dir = $collectd::params::plugin_conf_dir
   validate_bool($purge_config, $fqdnlookup)
   validate_array($include, $typesdb)
+
+  # Version for templates
+  $collectd_version = pick(
+    $::collectd_real_version,                                                      # Fact takes precedence
+    regsubst(
+      regsubst($version,'^(absent|held|installed|latest|present|purged)$', ''), # standard package resource ensure value? - strip and return undef
+      '^\d+(?:\.\d+){1.2}', '\0'),                                                 # specific package version? return only semantic version parts
+    $minimum_version,
+    '1.0')
 
   package { $package_name:
     ensure   => $version,
--- a/lib/facter/collectd_real_version.rb
+++ b/lib/facter/collectd_real_version.rb
@@ -1,4 +1,4 @@
-# Fact: collectd_version
+# Fact: collectd_real_version
 #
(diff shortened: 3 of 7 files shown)


blkperl:
PIng @txaj for review

blkperl:
Is it possible to write a spec test to test the new behavior?

bdellegrazie:
Hi @blkperl, I'd like to add tests for the new functionality but I'm very new to puppet-rspec and I'm not sure how. Specifically I'm having difficulty accessing the `$_collectd_version` created in `init.pp` as its essentially being used as a module-wide variable in the templates.
Do you have any suggestions?

bdellegrazie:
ping @blkperl and @txaj - any update or further comments?

README.md
```diff
@@ -1087,7 +1092,10 @@ See metadata.json for supported platforms
 
 ##Known issues
 
-Some plugins will need two runs of Puppet to fully generate the configuration for collectd. See [this issue](https://github.com/puppet-community/puppet-collectd/issues/162).
+###Puppet needs two runs to correctly write my conf, why?
+
+Some plugins will need two runs of Puppet to fully generate the configuration for collectd. See [this issue](https://github.com/pdxcat/puppet-module-collectd/issues/162).
+This can be avoided by specifying an explicit version (`$version`) or a minimum version (`$minimum_version`) to the collectd class
```
txaj: Could you be more explicit about how this parameter works ? like "setting this parameter to 9.99 will makes this module assume at the first run (when the fact responsible to provide the collectd version is not yet avaibable) that your systems are running collectd 9.99, and generate the configuration accordingly"

manifests/init.pp
```diff
@@ -14,12 +14,16 @@
...
   $version                = installed,
+  $minimum_version        = undef,
 ) inherits collectd::params {
 
   $plugin_conf_dir = $collectd::params::plugin_conf_dir
   validate_bool($purge_config, $fqdnlookup)
   validate_array($include, $typesdb)
 
+  # Version for templates
+  $_collectd_version = pick($::collectd_version, regsubst($version, '^\d+(?:\.\d+){1.2}', '\0'), $minimum_version)
```
txaj: what is this regex for ? if that's for sanity, I'd prefer a validate_re so we can fail if the input is incorrect

txaj: this changelog hurts. Instead of changing collectd_version everywhere, why not change the name of the fact ? ping @blkperl do you see any argument to say that this could be a bad idea ?

txaj:
@bdellegrazie for testing I only see one way : take a module that generates a different configuration depending on the collectd version, and test the output when you set the fact to undef/value1/value2 and the default to undef/value1/value2 and check that it makes sense

README.md
bdellegrazie: Sure will do

manifests/init.pp
bdellegrazie: The regex is only meant to match a version number of collectd (as much of the major, minor and patch parts that are supplied). it returns undef if version happens to be 'installed', 'present' or 'absent'. validate_re is not appropriate in this circumstance.

re: fact name
I don't mind changing the fact name if that's what is preferred. I changed the templates because I was concerned that the fact was considered a part of the modules 'public' interface.
Make a decision either way and let me know.

bdellegrazie:
@txaj, @blkperl: I've updated the PR to do the following:
* renamed collectd_version fact to collectd_real_version
* corrected an error with the regexp which selects the version for the templates.

I had to use a test module in the fixtures dir to be able to create the tests around the collect_version but they do work.

igalic:
thanks @bdellegrazie and @txaj! i've merged this manually.

igalic:
since merging this, we now have a [failing test](https://github.com/puppet-community/puppet-collectd/blob/master/spec/classes/collectd_init_spec.rb#L73-L77)

@bdellegrazie do i understand it correctly, that we should be checking for collectd_real_version here (as well as in the code)?

bdellegrazie:
yes the fact should be collectd_real_version - I thought it was on my
branch?

On 28 September 2015 at 16:15, Igor Galić <notifications@github.com> wrote:

> since merging this, we now have a failing test
> <https://github.com/puppet-community/puppet-collectd/blob/master/spec/classes/collectd_init_spec.rb#L73-L77>
>
> @bdellegrazie <https://github.com/bdellegrazie> do i understand it
> correctly, that we should be checking for collectd_real_version here (as
> well as in the code)?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/puppet-community/puppet-collectd/pull/305#issuecomment-143773346>
> .
>



-- 
Kind regards,

Brett Delle Grazie

bdellegrazie:
Okay, I realise someone has added more tests (great!) but when merging the
fact needs to be collectd_real_version - then the internal
'collectd_version' variable that is used by the templates will resolve
correctly.

On 28 September 2015 at 16:41, Brett Delle Grazie <
brett.dellegrazie@gmail.com> wrote:

> yes the fact should be collectd_real_version - I thought it was on my
> branch?
>
> On 28 September 2015 at 16:15, Igor Galić <notifications@github.com>
> wrote:
>
>> since merging this, we now have a failing test
>> <https://github.com/puppet-community/puppet-collectd/blob/master/spec/classes/collectd_init_spec.rb#L73-L77>
>>
>> @bdellegrazie <https://github.com/bdellegrazie> do i understand it
>> correctly, that we should be checking for collectd_real_version here (as
>> well as in the code)?
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/puppet-community/puppet-collectd/pull/305#issuecomment-143773346>
>> .
>>
>
>
>
> --
> Kind regards,
>
> Brett Delle Grazie
>



-- 
Kind regards,

Brett Delle Grazie

igalic:
@bdellegrazie did i do a poor job of rebasing / merging your branch?

bdellegrazie:
@igalic I'd hardly count two functions as a poor merge :)
No, there is however something 'broken' with Puppet 4.0 according to Travis output from PR #343.
I don't know how to fix this one however.

PR #311, opened 2015-08-06
Add rubocop syntax checking

blkperl:


blkperl:
@daenney good idea or crazy?

daenney:
I would totally Rubocop but for the love of god and all that is holy don't commit your Gemfile.lock.

nibalizer:
@daenney THANK YOU @blkperl why is this so hard?

nibalizer:
Let's use https://github.com/puppetlabs/puppet/blob/master/.rubocop.yml as our rules.

daenney:
I agree, pick the Puppetlabs one and stay in sync with it. Means the code should have a similar look/feel and behaviour between community T&P's and stuff shipped with Puppet.

blkperl:
@daenney ok

PR #318, opened 2015-08-11
Add support for PrintSeverity option (fix #317)

### LINKED ISSUE #317: Cannot set PrintSeverity with collectd::plugin::logfile
See https://collectd.org/documentation/manpages/collectd.conf.5.shtml#plugin_logfile

```
PrintSeverity true|false

    When enabled, all lines are prefixed by the severity of the log message, for example "warning". Defaults to false.
```

According to collectd's changelog, this has been added in 4.10.0.

simonpasquier:


blkperl:
Thanks for the contribution

PR #320, opened 2015-08-13
Small fix in documentation.

ttarczynski:
Two sections were mixed with each other:
* Class: collectd::plugin::processes
* Class: collectd::plugin::protocols


blkperl:
Thanks for the contribution!

PR #321, opened 2015-08-14
Enhance and fix snmp plugin

ortegaga:
Added support for InstancePrefix, Scale and Shift options, both when using the parametized class and the defined type.

Also it's removed the sorting of Values passed to parametized class, because you can feed the wrong values to the type defined in types.db (in fact, they are not sorted when using the defined type).

Finally, check whether the option Table is set when using the parametized class


blkperl:
Thanks for the contribution!

PR #324, opened 2015-08-19
Add innodbstats and slavenotifications for mysql

t-8ch:


jlory:
You should check if innodbstats & slavenotifications  variable are set to true before appending to the template otherwise people using Collectd < 5.5 will have InnodbStats = false in their config file.

t-8ch:
@jlory I made it completely backwards compatible.

jlory:
@t-8ch Awesome, than you!

templates/mysql-database.conf.erb
```diff
@@ -15,5 +15,11 @@
 <%- if @socket -%>
     Socket "<%= @socket %>"
 <%- end -%>
+<%- if not @innodbstats.nil? -%>
+    InnodbStats <%= @innodbstats %>
+<%- end -%>
+<%- if not @slavenotifications.nil? -%>
```
blkperl: You can just do `if @slavenotifications`

templates/mysql-database.conf.erb
```diff
@@ -15,5 +15,11 @@
 <%- if @socket -%>
     Socket "<%= @socket %>"
 <%- end -%>
+<%- if not @innodbstats.nil? -%>
```
blkperl: You can do `if @innodbstats`  instead

templates/mysql-database.conf.erb
t-8ch: I wanted to preserve an explicit `slavenotification = false`

blkperl:
Thanks for the contribution!

PR #326, opened 2015-08-21
allow_user_group_settings

arenstar:


danzilio:
It looks like we're making some assumptions here that are not aligning with reality:

```
Notice: Compiled catalog for ubuntu in environment production in 0.41 seconds
Info: Applying configuration version '1440635452'
Notice: /Stage[main]/Collectd/Package[collectd]/ensure: ensure changed 'purged' to 'present'
Notice: /Stage[main]/Collectd/File_line[include_conf_d_dot_conf]/ensure: created
Info: /Stage[main]/Collectd/File_line[include_conf_d_dot_conf]: Scheduling refresh of Service[collectd]
Error: Could not set 'directory' on ensure: Could not find user collectd at 41:/etc/puppet/modules/collectd/manifests/init.pp
Error: Could not set 'directory' on ensure: Could not find user collectd at 41:/etc/puppet/modules/collectd/manifests/init.pp
Wrapped exception:
Could not find user collectd
Error: /Stage[main]/Collectd/File[collectd.d]/ensure: change from absent to directory failed: Could not set 'directory' on ensure: Could not find user collectd at 41:/etc/puppet/modules/collectd/manifests/init.pp
Notice: /Stage[main]/Collectd/Service[collectd]: Dependency File[collectd.d] has failures: true
Warning: /Stage[main]/Collectd/Service[collectd]: Skipping because of failed dependencies
Notice: /Stage[main]/Collectd/Service[collectd]: Triggered 'refresh' from 1 events
Info: Creating state file /var/lib/puppet/state/state.yaml
Notice: Finished catalog run in 12.67 seconds
```

This is on an Ubuntu 14.04.1 box. Log here: http://ci.puppet.community/buildlogs/puppet-community+puppet-collectd+326+1440635475+FAIL

Basically, this module doesn't manage the user, so we're running into a condition where Puppet is trying to create a directory with an owner that doesn't exist yet because the package hasn't been installed. We can opt to manage the user in this module (meh), or do some better dependency management.

Could we also change `root_user` to `collectd_user` or `user` or something like that? In my mind, `root_user` will always be `root`.

blkperl:
Please reopen when the tests pass and its mergable.

PR #327, opened 2015-08-25
Add support for the new AllPortsSummary option.

dfilion:
Add support for the allportssummary option added to the tcpconns module introduced in version collectd-5.5.0.


blkperl:
Thanks for the contribution!

PR #328, opened 2015-09-01
Factor our resources into classes

danzilio:
This is the first in a string of PRs to move the plugins into their own
module. This PR pulls the resources associated with the
package-file-service paradigm and moves them into classes. This way we
can have a cleaner interface between modules. Instead of a resource
notifying 'Service[collectd]' we can send a refresh signal to
'Class[collectd::service]'.

This also factors out the various dependencies and manages them at the
class level.


danzilio:
This also adds parameters for the things that were hardcoded in, and looked up from the `params` class.

igalic:
i *like* this change. a lot

PR #329, opened 2015-09-01
Added plugin::netlink

ttarczynski:
I need the netlink plugin to get detailed network interface stats:
- https://collectd.org/wiki/index.php/Plugin:Netlink

I've based it on `collectd::plugin::interface`
I will add short docs and will try to write rspec tests.


blkperl:
Thanks for the contribution!

PR #335, opened 2015-09-10
Added plugin write_sensu

waldosCH:
It would be a nice to have the plugin write sensu availlable.
https://collectd.org/wiki/index.php/Plugin:Write_Sensu
Thanks a lot in advance for your comments and inputs.


blkperl:
Awesome! Thanks for the contribution

PR #341, opened 2015-09-22
Make the management of the collectd package optional

jovandeginste:
Sometimes installing packages is done in a different way (eg. epel may
not be enabled by default)

The dependencies still hold.

### DIFF
--- a/manifests/plugin/amqp.pp
+++ b/manifests/plugin/amqp.pp
@@ -1,6 +1,7 @@
 # https://collectd.org/wiki/index.php/Plugin:AMQP
 class collectd::plugin::amqp (
   $ensure          = present,
+  $manage_package  = $true,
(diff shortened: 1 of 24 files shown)


blkperl:
This doesn't merge cleanly

jovandeginste:
Rebased (and updated) the changes

bastelfreak:
@jovandeginste thanks for contributing!

manifests/plugin/amqp.pp
```diff
@@ -1,6 +1,7 @@
 # https://collectd.org/wiki/index.php/Plugin:AMQP
 class collectd::plugin::amqp (
   $ensure          = present,
+  $manage_package  = $true,
```
igalic: this should be collectd::params::manage_package

PR #344, opened 2015-09-29
fix class tests

igalic:
this fixes #343


igalic:
thanks @bdellegrazie!

PR #345, opened 2015-09-30
Fixing duplicate param and some housekeeping

danzilio:
Hauskeeping! This commit removes a duplicate $version paramter. It
also moves the $minimum_version default to the params class and
aligns an inline comment.


daenney:
:100:

PR #346, opened 2015-10-01
Revert "fix class tests"

igalic:
Reverts puppet-community/puppet-collectd#344 as per @bdellegrazie's recommendation


igalic:
a better fix is in #347

PR #353, opened 2015-10-06
Update from puppet-community modulesync_configs

igalic:


blkperl:
Thanks for the contribution! :dancer:

PR #357, opened 2015-10-12
Package install options

ghoneycutt:


manifests/init.pp
```diff
@@ -1,29 +1,30 @@
...
+  $purge_config            = $collectd::params::purge_config,
+  $recurse                 = $collectd::params::recurse,
+  $threads                 = $collectd::params::threads,
+  $timeout                 = $collectd::params::timeout,
+  $typesdb                 = $collectd::params::typesdb,
+  $write_queue_limit_high  = $collectd::params::write_queue_limit_high,
+  $write_queue_limit_low   = $collectd::params::write_queue_limit_low,
+  $config_file             = $collectd::params::config_file,
+  $package_provider        = $collectd::params::provider,
+  $package_install_options = $collectd::params::package_install_options,
```
ghoneycutt: Lots of noise to line these back up. I simply added `$package_install_options`.

manifests/install.pp
```diff
@@ -1,16 +1,21 @@
 class collectd::install (
-  $version          = $collectd::version,
-  $package_name     = $collectd::package_name,
-  $package_provider = $collectd::package_provider,
-  $manage_package   = $collectd::manage_package,
+  $version                 = $collectd::version,
+  $package_name            = $collectd::package_name,
+  $package_provider        = $collectd::package_provider,
+  $package_install_options = $collectd::params::package_install_options,
```
ghoneycutt: Lots of noise to line these back up. I simply added `$package_install_options`.

manifests/params.pp
```diff
@@ -1,24 +1,25 @@
...
+  $timeout                 = 2
+  $typesdb                 = []
+  $write_queue_limit_high  = undef
+  $write_queue_limit_low   = undef
+  $version                 = installed
+  $service_ensure          = running
+  $service_enable          = true
+  $minimum_version         = undef
+  $manage_package          = true
+  $package_install_options = undef
```
ghoneycutt: Lots of noise to line these back up. I simply added `$package_install_options`.

manifests/plugin/apache.pp
```diff
@@ -1,17 +1,23 @@
...
 class collectd::plugin::apache (
-  $ensure         = present,
-  $manage_package = $true,
-  $instances      = { 'localhost' => { 'url' => 'http://localhost/mod_status?auto' } },
-  $interval       = undef,
+  $ensure                  = present,
+  $manage_package          = $true,
+  $instances               = { 'localhost' => { 'url' => 'http://localhost/mod_status?auto' } },
+  $interval                = undef,
+  $package_install_options = $collectd::params::package_install_options,
```
danzilio: My one question here is: what's our interface? Should we use `$collectd::package_install_options` for this? What would the user expect here? I'm not sure, that's why I'm posing the question.

ghoneycutt: You are correct, we should use `$collectd::package_install_options`. If you need to specify `install_options` it is very likely you would need it for everything, so defaulting to that is a good idea.

ghoneycutt:
@danzilio Thanks for the help!

PR #358, opened 2015-10-18
Mongodb plugin

sharlander:
adds a python based mongodb plugin

### DIFF
--- a/manifests/plugin/mongodb.pp
+++ b/manifests/plugin/mongodb.pp
@@ -39,28 +39,6 @@ class collectd::plugin::mongodb (
     }
   }
 
-  if $ensure == 'present' {
-    file { 'collectd_modulepath':
-      ensure  => 'directory',
-      path    => $collectd_dir,
-      owner   => 'root',
-      group   => 'root',
-      mode    => '0755',
-      require => Package['collectd-python'],
-    }
-  }
-
-  file { 'mongodb.py':
-    ensure  => $ensure,
-    path    => "${collectd_dir}/mongodb.py",
-    source  => 'puppet:///modules/collectd/mongodb.py',
-    owner   => 'root',
-    group   => 'root',
-    mode    => '0644',
-    require => File['collectd_modulepath'],
-    notify  => Service['collectd'],
-  }
-
   collectd::plugin { 'mongodb':
     ensure   => $ensure,
     content  => template('collectd/plugin/mongodb.conf.erb'),
--- a/manifests/plugin/mongodb.pp
+++ b/manifests/plugin/mongodb.pp
@@ -3,7 +3,6 @@
 class collectd::plugin::mongodb (
   $ensure         = 'present',
   $interval       = undef,
-  $manage_package = true,
   $db_host        = '127.0.0.1',
   $db_user        = undef,
   $db_pass        = undef,
@@ -30,19 +29,9 @@ class collectd::plugin::mongodb (
     fail("collectd::plugin::mongodb::db_pass is <undef>, please specify the password for db user: ${db_user}")
   }
 
-  if $::osfamily == 'Redhat' {
-    if $manage_package == true {
-      package { 'collectd-python':
-        ensure          => $ensure,
-        install_options => ['--nogpgcheck'],
-      }
-    }
-  }
-
   collectd::plugin { 'mongodb':
     ensure   => $ensure,
     content  => template('collectd/plugin/mongodb.conf.erb'),
     interval => $interval,
-    require  => File['mongodb.py'],
   }
 }
--- a/manifests/plugin/mongodb.pp
+++ b/manifests/plugin/mongodb.pp
@@ -1,8 +1,9 @@
-# == Class: collectd::plugin::mongodb
-#
-# install and configure mongodb plugin
+# Class: collectd::plugin::mongodb
 #
 class collectd::plugin::mongodb (
+  $ensure         = 'present',
+  $interval       = undef,
+  $manage_package = true,
   $db_host        = '127.0.0.1',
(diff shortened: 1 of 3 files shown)


igalic:
@sharlander i'm very confused about `files/mongodb.py`.
where does it come from?
is it your own implementation?
why are you submitting it here, and not upstream?


what is the canonical way of installing third-party plugins?
etc…

manifests/plugin/mongodb.pp
```diff
@@ -0,0 +1,70 @@
...
+      group   => 'root',
+      mode    => '0755',
+      require => Package['collectd-python'],
+    }
+  }
+
+  file { 'mongodb.py':
+    ensure  => $ensure,
+    path    => "${collectd_dir}/mongodb.py",
+    source  => 'puppet:///modules/collectd/mongodb.py',
```
ghoneycutt: @sharlander could this please go in files/plugins/mongodb.py ?

ghoneycutt:
@igalic we inherited a fork with local additions and would like to get them merged. We would like the python script to be managed here as putting in into some custom package that this module would install would make it extremely hard for anyone to actually consume.

@sharlander could you please rebase and squash your commits?

ffrank:
I'm with @igalic. If the plugin is in working condition, you should seek to add it to mainline `collectd`.

Failing that, I kind of feel that there should be an optional module with collectd plugins, with clear instructions for the user how to install and use them. The `collectd` module itself should focus on the tool itself.

igalic:
thank you for your comments, @ffrank.

i've had a bit of a discussion on #puppet-community, and here are some of my thoughts:

* this is a (third-party) collectd plugin (using the binding-API, so it's not directly *in* collectd)
* this plugin could be interesting to *other* communities, not just ourselves (chef, ansible, and other friends!)
* collectd does not offer a way to install third-party plugins, however
* puppet-collectd should offer a way to install third-party plugins

as such, we should

* extract this plugin from `files`, and put it in its own repository, perhaps @puppet-community/collectd-mongodb
* write a (general enough) deployer (using package, vcsrepo, file, archive, depending on what people need)

n.b.: This approach will also hold true for mcollective, which can be extended with plugins, but so far only provides packages to install them

and while I am perfectly aware of [fpm](https://github.com/jordansissel/fpm), not everyone out there is comfortable running a packaging infrastructure for such ~trivial things~

(yes, i *know* of [packagecloud](https://packagecloud.io/))

sharlander:
@igalic @ffrank @ghoneycutt sorry for the late response, i been really busy this week, anyway...

i see there was a lot of discussion going on about this, so in the end the conclusion for me is,
we will need to write a third party plugin deployment for collectd and add the python file to a separate module/repo
so i will close this pull request and move on to write a generic plugin deployment

ffrank:
Well, it's fine to manage the configuration for the plugin here, right?

So let's please keep the PR, just remove the custom collectd plugin.

sharlander:
@ffrank yes, you're right
i rebased the branch and removed the file resources

manifests/plugin/mongodb.pp
```diff
@@ -0,0 +1,48 @@
...
+  }
+  elsif $db_pass == undef {
+    fail("collectd::plugin::mongodb::db_pass is <undef>, please specify the password for db user: ${db_user}")
+  }
+
+  if $::osfamily == 'Redhat' {
+    if $manage_package == true {
+      package { 'collectd-python':
+        ensure          => $ensure,
+        install_options => ['--nogpgcheck'],
```
bastelfreak: * Which repo owns this package?
* Why do we need to pass this option?
* sounds like a general package and not mongodb releated, should this be moved to another class?

manifests/plugin/mongodb.pp
```diff
@@ -0,0 +1,48 @@
...
+      }
+    }
+  }
+
+  collectd::plugin { 'mongodb':
+    ensure   => $ensure,
+    content  => template('collectd/plugin/mongodb.conf.erb'),
+    interval => $interval,
+    require  => File['mongodb.py'],
```
bastelfreak: the require is useless since you removed the file resource

sharlander:
@bastelfreak oh damn, i should have paid more attention when i removed the file resources

regarding the package i thought it's kind of neccesary to run the python plugin, but on second thought i agree with you it should be rather handled via rpm dependency or the plugin deployment module,
so i removed that now

bastelfreak:
well, is it necessary to run the plugin or isn't it? If it's needed please extend the module to install it (maybe the params.pp is a good point to start)

ffrank:
@bastelfreak the plugin used to be part of the PR, but we asked @sharlander to yank that out. You are right for pointing out that the dependency package should go away as well.

sharlander:
@bastelfreak @ffrank  i removed the package resource, as mentioned above the package can be installed somewhere else, so i guess this should be fine now

ghoneycutt:
FYI: The python file in question that was removed from the commit is a fork of https://github.com/sebest/collectd-mongodb/blob/master/mongodb.py

igalic:
@ghoneycutt cool. so that's bsd licensed, so at least it's compatible with what we did with it ;)

nibalizer:
Can we squash this please? Also add a test to specs/classes/ to be in-line with the rest of the classes?

ghoneycutt:
@sharlander take a look at https://github.com/puppet-community/puppet-collectd/blob/master/spec/classes/collectd_plugin_dns_spec.rb for how to do the spec tests

@nibalizer Thank you!

ghoneycutt:
@sharlander Please close this. Worked with @jonmccreery in PR #371 to add spec tests

Issue #355, opened 2015-10-06
MySQL master/slave stats


mookie-:
Hey,

I think you should remove this code from `manifests/plugin/mysql/database.pp`:

```
  if ($masterstats == true and $slavestats == true) {
    fail('master and slave statistics are mutually exclusive.')
  }
```

MySQL supports Master/Master replication where each Master is also a slave and you probably want both. It's possible with the collectd plugin so I don't think you should disable this here.

igalic:
fixed in #365 by @blacktoko

PR #359, opened 2015-10-18
Removed if statement to check if the master and slave stats are both true

blacktoko:
Fixes: #355  MySQL supports Master/Master replication where each Master is also a slave and you probably want both.

### DIFF
--- a/manifests/plugin/mysql/database.pp
+++ b/manifests/plugin/mysql/database.pp
@@ -31,10 +31,6 @@ define collectd::plugin::mysql::database (
     validate_bool($slavenotifications)
   }
 
-  if ($masterstats == true and $slavestats == true) {
-    fail('master and slave statistics are mutually exclusive.')
-  }
-
   file { "${name}.conf":
     ensure  => $ensure,
     path    => "${conf_dir}/mysql-${name}.conf",
--- a/manifests/plugin/mysql/database.pp
+++ b/manifests/plugin/mysql/database.pp
@@ -6,6 +6,8 @@ define collectd::plugin::mysql::database (
   $username           = 'UNSET',
   $password           = 'UNSET',
   $port               = '3306',
+  $masterstats        = false,		
+  $slavestats         = false,
   $socket             = undef,
   $innodbstats        = undef,
   $slavenotifications = undef,
@@ -16,7 +18,8 @@ define collectd::plugin::mysql::database (
   $conf_dir = $collectd::params::plugin_conf_dir
 
   validate_string($database, $host, $username, $password, $port)
-
+  validate_bool($masterstats, $slavestats)
+  
   if $socket {
     validate_string($socket)
   }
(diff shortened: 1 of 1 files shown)


blacktoko:
The centos7 test fails, but it looks like it has nothing to do with the change i made. The installation of the centos machine goes wrong, at first it seems the connection trough ssh failes. And after that the installation of some ruby gems failes. The error seems to come from this file on line 26:
```
puppetlabs-puppetdb/spec/spec_helper_acceptance.rb
```

blacktoko:
p.s. this is my first pull request, i hope i did it all right :-)

ffrank:
Hi, thanks for the PR.

I concur that for MySQL itself, it makes no sense to define master/slave as mutually exclusive. However, it must be positively determined that the *collectd* plugin will accept this state. The [documentation](https://collectd.org/wiki/index.php/Plugin:MySQL) appears inconclusive to me.

Furthermore, the parameters cannot just be removed. Be mindful of the template that is invoked by the altered class and makes [use](https://github.com/puppet-community/puppet-collectd/blob/master/templates/mysql-database.conf.erb#L13-L14) of those values. Yes, they weren't just for show ;-)

igalic:
yes, @ffrank, but, what does the plugin's source code say? :P

blacktoko:
I brought back the paramaters in the last commit. The only change is now the removal of the if statement.

manifests/plugin/mysql/database.pp
```diff
@@ -6,7 +6,7 @@
   $username           = 'UNSET',
   $password           = 'UNSET',
   $port               = '3306',
-  $masterstats        = false,
+  $masterstats        = false,
```
igalic: that trailing space needs to go

igalic:
@blacktoko i'd also appreciate if all of those three commits were [squashed down](https://www.git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Squashing-Commits) to one

ffrank:
@blacktoko can we help you with the branch maintenance?

blacktoko:
@ffrank i did manage to remove my old patches and created one good new one. But i need to learn myself some git basics. For example: I could not manage to create a pullrequest from the commandline. So i did that trough the web interface. I'll find a basic git tutorial so i can be of more help :-) Thanks for the offer!

ffrank:
Cool, thanks for the feedback.

Hint: You *cannot* create a PR from the console. You push to your fork, then visit the web interface (your fork has a handy link to the latest branch highlighted) and open the PR. So I think you did it right.

It is always best to keep working with the original branch, and change it (in your fork) using `git push -f`. The existing PR reflects these changes then.

igalic:
@blacktoko for more hand-holding, i'm available during vienna office-hours in #puppet-community, #puppet-dev, and #puppet on Freenode :)

blacktoko:
Thanks for all your help! I'll join those channels, i'm known as blacktoko there too.

PR #360, opened 2015-10-19
Solaris variables

t-8ch:
# Fix service name

With the current name I get:

```sh
root@host:~# /usr/sbin/svcadm enable -s collectd
svcadm: Pattern 'collectd' doesn't match any instances
```

# Fix the config file location

The current settings are equivalent to dropping a directory named 'conf.d' into
'/etc' which makes in nonobvious to which package it belongs.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -38,7 +38,7 @@ class collectd::params {
       $provider          = 'pkgutil'
       $collectd_dir      = '/etc/opt/csw'
       $plugin_conf_dir   = "${collectd_dir}/conf.d"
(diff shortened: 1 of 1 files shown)


ffrank:
Hi, thanks for this PR.

Can this possibly change behavior for existing users?

igalic:
@ffrank yes: it should make things work for current users.

t-8ch:
@igalic It can also change behaviour if users hardcoded the old locations into their custom resources.
The alternatives would be:
* Create a symlink from/to the old location
* Make the directory configurable

PR #361, opened 2015-10-20
[filecount] do not an emit empty <Plugin> block

t-8ch:
Our collectd 4.1 chokes on an empty Plugin block.
This can happen when not using the `directories` parameter to
collectd::plugin::filecount and instead using only
collectd::plugin::filecount::directory .

### DIFF
--- a/spec/classes/collectd_plugin_filecount_spec.rb
+++ b/spec/classes/collectd_plugin_filecount_spec.rb
@@ -57,4 +57,11 @@ describe 'collectd::plugin::filecount', :type => :class do
       should compile.and_raise_error(/String/)
     end
   end
+
+  context ':directories is empty' do
+    it 'Will not create an empty <Plugin "filecount"> block' do
+      should contain_file('filecount.load')
+        .without_content(/<Plugin "filecount">/)
+    end
+  end
 end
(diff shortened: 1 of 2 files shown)


templates/plugin/filecount.conf.erb
```diff
@@ -1,3 +1,4 @@
+<%-if not @directories.empty? -%>
```
igalic: `if not` → `unless` please

t-8ch: done

igalic:
@t-8ch could you also extend this spec test, please? https://github.com/puppet-community/puppet-collectd/blob/master/spec/defines/collectd_plugin_filecount_directory_spec.rb

t-8ch:
@igalic Not https://github.com/puppet-community/puppet-collectd/blob/master/spec/classes/collectd_plugin_filecount_spec.rb?

spec/classes/collectd_plugin_filecount_spec.rb
```diff
@@ -57,4 +57,11 @@
       should compile.and_raise_error(/String/)
     end
   end
+
+  context ':directories is empty' do
+    it 'Will not create an empty <Plugin "filecount"> block' do
+      should contain_collectd__plugin('filecount')
+        .with_content(/\s*/)
```
igalic: perhaps a better match would be: `.without_content(/<Plugin "filecount">/)` ?

t-8ch: @igalic done

igalic:
thanks @t-8ch!

PR #362, opened 2015-10-22
collectd < 5.5 doesn't come with mysql InnodbStats.

bzed:
The InnodbStats option of the mysql plugin was introduced in collectd 5.5.

### DIFF
--- a/templates/mysql-database.conf.erb
+++ b/templates/mysql-database.conf.erb
@@ -15,8 +15,10 @@
 <%- if @socket -%>
     Socket "<%= @socket %>"
 <%- end -%>
+<% if @collectd_version and (scope.function_versioncmp([@collectd_version, '5.5']) >= 0) -%>
 <%- if not @innodbstats.nil? -%>
     InnodbStats <%= @innodbstats %>
+<%- end -%>
 <%- end -%>
 <%- if not @slavenotifications.nil? -%>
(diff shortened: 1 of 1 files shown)


igalic:
shouldn't we fail, or `warn()` from within the manifest, rather than ignore this option in the template?

ffrank:
If we `warn` (with which I would be happy), we will also want to `notify { "...": loglevel => "warning" }` (or is that "warn" as well?)

We actually want the `notify` more, for that will definitely show up in reports.

igalic:
do elaborate, please.

ffrank:
Invoking the `warn` function on the master will leave a Warning level message in the master log. I'm not sure whether reporting reflects that. I would assume not, since the agent is responsible for compiling the report after its run.

On the other hand, you can do this:

    notify { 'You are doing something insensible':
      loglevel => 'warning',
    }

This shows up in a terrible color on the console when running interactive, and it *will* light up the reports. (I did manage to look up the available [levels](https://docs.puppetlabs.com/references/latest/metaparameter.html#loglevel) by now.)

igalic:
ack, thanks for the clarification.

jyaworski:
Thanks for your contribution!

@igalic why is PCCI voting failing on a lot of PRs?

jyaworski:
I'm going to make an executive decision and merge this.

Thanks!

PR #363, opened 2015-10-23
added manifest and template to support smart plugin

devfaz:
Hi,

just wrote a small smart-plugin.

 Fabian

### DIFF
--- a/manifests/plugin/smart.pp
+++ b/manifests/plugin/smart.pp
@@ -1,0 +1,27 @@
+#== Class: collectd::plugin::smart
+#
+# Class to manage smart plugin for collectd 
+# === Parameters
+# [*ensure*]
+#   ensure param for collectd::plugin type
+# 
+# [*disks*]
+#   array of disks to create config for 
+#   example: ['sda', 'sdb', 'sdc']
+#
+# [*ignoreselected*]
+#   If enabled, ignore given disks
+#
+class collectd::plugin::smart (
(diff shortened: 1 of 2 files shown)


ffrank:
That looks neat, thanks.

Please take note of the [discussion](https://github.com/puppet-community/puppet-collectd/pull/358#issuecomment-150177836) that is currently going on over in #358. We may want to manage this kind of external plugin outside of this module in the future.

igalic:
@ffrank i think we should post $discussion on $mailinglist.

blkperl:
Hi, Thanks for the contribution
 
Please add docs to the README and spec tests.

Also I noticed their are no docs on the official wiki, if you know how to configure the plugin you may want to contribute to the wiki: https://collectd.org/wiki/index.php/Plugin:smart

daenney:
The smart plugin is part of collectd since 5.5, it's not an external plugin so that discussion doesn't really apply.

https://collectd.org/documentation/manpages/collectd.conf.5.shtml#plugin_smart

ffrank:
D'oh, if I could have been bothered to actually look at the change, I would have noticed that it does not in fact include collectd plugin code.

Thanks @daenney. Also +1 to what @blkperl says above. @devfaz if you need help with tests or docs, please let us know.

blkperl:
Closing due to lack of activity. Please reopen when the feedback has been addressed

PR #364, opened 2015-10-23
collectd may fail to start in wheezy if processes-plugin is used ...

devfaz:
collectd may fail to start in wheezy if proccesses-plugin is used without any processes.

      root@puppetclient:/etc/collectd/conf.d# collectd -f
      Parse error in file `/etc/collectd/conf.d/processes-config.conf', line 104 near `/': syntax error, unexpected SLASH, expecting UNQUOTED_STRING
      yyparse returned error #1 
      configfile: Cannot read file `/etc/collectd/conf.d/processes-config.conf'.

Problem: creating a processes-config.conf without any configured processes is unnecessary and leads to above error (at least in wheezy).

Solution: create a processes-config.conf only if there are processes configured.

Thanks,

Fabian

### DIFF
--- a/manifests/plugin/processes.pp
+++ b/manifests/plugin/processes.pp
@@ -17,8 +17,14 @@ class collectd::plugin::processes (
     interval => $interval,
   }
 
+  if ( $processes or $process_matches ) {
+    $process_config_ensure = present
+  } else {
+    $process_config_ensure = absent
+  }
+
   concat{"${collectd::params::plugin_conf_dir}/processes-config.conf":
-    ensure         => $ensure,
+    ensure         => $process_config_ensure,
     mode           => '0640',
     owner          => 'root',
     group          => $collectd::params::root_group,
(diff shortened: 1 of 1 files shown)


ffrank:
Thanks for the PR. I'm leaving a comment inline.

manifests/plugin/processes.pp
```diff
@@ -17,27 +17,33 @@
...
-  concat{"${collectd::params::plugin_conf_dir}/processes-config.conf":
-    ensure         => $ensure,
-    mode           => '0640',
-    owner          => 'root',
-    group          => $collectd::params::root_group,
-    notify         => Service['collectd'],
-    ensure_newline => true,
+  if ( $processes or $processes_matches ) {
+    concat{"${collectd::params::plugin_conf_dir}/processes-config.conf":
+      ensure         => $ensure,
```
ffrank: Would it not be simpler, in order to avoid this huge if block, to use an alternate `$ensure` value that is derived from the condition?

    if ( $processes or $processes_matches ) {
      $process_config_ensure = true
    } else { ...

devfaz:
Great idea, just updated my branch.

Fabian

manifests/plugin/processes.pp
```diff
@@ -17,8 +17,14 @@
     interval => $interval,
   }
 
+  if ( $processes or $processes_matches ) {
```
igalic: this is giving us the error:
```
manifests/plugin/processes.pp:20:variable_scope:WARNING:top-scope variable being used without an explicit namespace
```

which is confusing… did you typo?
doesn't look like a typo to me :S

ffrank: It is though. It's just `$process_matches`.

ffrank:
Much more palatable, thanks @devfaz.

When you fix the typo - can you please squash all commits into one?

Let us know if you need help with that.

devfaz:
Fixed :)

igalic:
thanks @devfaz!

PR #365, opened 2015-10-23
#355 Removed Master/Slave if statement

blacktoko:
#355 removed if statement becouse MySQL supports Master/Master replication where each Master is also a slave and you probably want both.


blacktoko:
I removed the old pullrequest and did it over. Could not get it right from the command line sorry! Have to practise more with this :-)

igalic:
thank you so much @blacktoko!

PR #367, opened 2015-11-03
removed group naming from regex for compatibility

hdoedens:
Naming a regex group within the regex itself using the '?<>' syntax only works on ruby >= 1.9.2.
I have removed the group naming.
Also escaped the dot in the url.


hdoedens:
After a closer look, ruby 1.8.7 is no longer supported. Which means instead of fixing the regex, I better upgrade ruby...

PR #369, opened 2015-11-10
fix $manage_package param

bastelfreak:
This fixes a bug that I introduced in https://github.com/puppet-community/puppet-collectd/pull/341 (wrong variable assignment)

### DIFF
--- a/manifests/plugin/amqp.pp
+++ b/manifests/plugin/amqp.pp
@@ -1,7 +1,7 @@
 # https://collectd.org/wiki/index.php/Plugin:AMQP
 class collectd::plugin::amqp (
   $ensure          = present,
-  $manage_package  = $true,
+  $manage_package  = $collectd::manage_package,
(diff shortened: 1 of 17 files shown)


danzilio:
:+1:

igalic:
uh…
what was `$true` before? `undef`??
so now we're fixing the issue, but breaking backwards compatibility?

daenney:
In a lot of plugins `$manage_package` doesn't actually do anything since they don't require external packages. In the other case it seems to only affect RHEL and what people probably did was explicitly pass in `true` or `false` because of the old broken behaviour.

I think this restores the expected behaviour so shouldn't be considered a backwards compatibility break.

bastelfreak:
I've got the same opinion as @daenney, this will restore the expected and shouldn't break anything, but I can try to add tests later

nibalizer:
Centos issue is known and will be fixed in #379

PR #371, opened 2015-11-13
Add a plugin for MongoDB

jonmccreery:


ghoneycutt:
@nibalizer Could you please merge this? @jonmccreery added spec tests, added documentation and squashed @sharlander 's commits as the previous PR asked for.

spec/classes/collectd_plugin_mongodb_spec.rb
```diff
@@ -0,0 +1,206 @@
...
+    context 'undefined' do
+      let :params do
+        { :db_pass => 'password',
+        }
+      end
+
+      it 'should fail' do
+        expect do
+          should contain_class('collectd::plugin::mongodb')
+        end.to raise_error(Puppet::Error, /is <undef> and must be a mongodb username/)
```
ffrank: Is this from the semantics of the mongodb plugin? With `mongodb` itself, it's (unfortunately) quite common to use no authentication, isn't it?

spec/classes/collectd_plugin_mongodb_spec.rb
```diff
@@ -0,0 +1,206 @@
...
+        default_params.merge(:configured_dbs => [25],
+                             :db_port        => '8080',)
+      end
+
+      dbport_single_fixture = File.read(fixtures('plugins/mongodb.conf.configured_dbs_single'))
+      it { should contain_file('mongodb.load').with_content(dbport_single_fixture) }
+    end
+  end
+
+  describe 'with configured_dbs parameter' do
```
ffrank: Style: this `describe` block shares the name with the previous one, I believe?

The `context` blocks could just as well share one `describe` block.

ffrank:
LGTM, except for two inline comments and a general remark about the tests:

There is quite a number of fixtures that can make many tests break whenever the template is changed, even if the change is trivial.

Could you change some (or all) of the tests to perform regex checks for the lines that actually pertain to the item under test, instead of the file content as a whole?

ghoneycutt:
@ffrank We used fixtures because the data is structured. If it was just lines a conf file, I would totally agree with using the regex, but since it is a quasi xml looking document, we need the fixtures.

ghoneycutt:
@jonmccreery please close in favor of PR #372 where I made the change requested by @ffrank to condense the duplicate describe blocks.

cc: @nibalizer

ffrank:
Thanks @ghoneycutt. I don't quite follow the reasoning regarding fixtures, especially seeing as the content of the `<Module mongodb>` tag is basically "lines in a config file".

No need to address this off the cuff. A second opinion might be helpful here.

I'd prefer if @jonmccreery could force-push the updated branch here, serial PRs make the history a little difficult to track. (Of course, if this is logistically challenging, we can go ahead w/ the new PR.)

ghoneycutt:
It is logistically challenging.

WRT fixtures, again the data is structured so checking for an individual line could pass, but if the line is not in the correct block, then it is actually a failure. You could get around this with putting the entire file in the spec test, but that is much harder to understand than using a fixture and is the problem that fixtures solve.

PR #374, opened 2015-11-20
templates: Fix version conditionals by using scope['::collectd::collectd_version'] instead of @collectd_version

caldwell:
@collectd_version is always nil due to 29dd796685469b3dc74a973eefca62ac1ed6e27f.


igalic:
i thought we had fixed that :|

jyaworski:
Hello.

Is this failing build? Can it be looked at?

bastelfreak:
@jyaworski will fix this in [PR393](https://github.com/voxpupuli/puppet-collectd/pull/393). We can close this one if his solution is merged.

jyaworski:
Fixed in #393

PR #376, opened 2015-12-09
fix collectd_real_version regex

azhurbilo:
fix collectd_real_version regex


igalic:
thank you so much for catching this

PR #378, opened 2015-12-15
Fix typo.

cstroe:
Which tests can be updated to fail because of this typo?


cstroe:
From the [build log](http://ci.puppet.community/buildlogs/puppet-community+puppet-collectd+378+1450199787+FAIL):

```
/home/pcci/new_ruby_gems_home/gems/beaker-2.14.1/lib/beaker/dsl/install_utils/foss_utils.rb:250:in `block in install_puppet_on': install_puppet() called for unsupported platform 'centos-7-1505-x86_64' on 'centos-7-1505-x86_64' (RuntimeError)
```

Is this something that I can fix?

cstroe:
Opened https://tickets.puppetlabs.com/browse/BKR-655 to fix the continuous-integration/pcci-centos7 failure.

igalic:
@cstroe that whole tests directory should probably be moved to examples.
but thank you so much for opening that beaker bug!

petems:
@cstroe I believe the beaker error is that it expects the following string for a host: `osname-releasenumber-arch`, so having the number there scares it

nibalizer:
This is a known issue with centos7 testing in pcci. Somone in #puppet said they would look at it.  I agree with @igalic that the whole file should move to examples//

PR #388, opened 2015-12-28
Update ceph.conf.erb to support boolean data type

vlitvin:
Currently ```<%= @longrunavglatency %>``` and ```<%= @convertspecialmetrictypes %>``` doesn't output anything, as they are boolean and not strings.
Monkey patch to fix that behavior.
Prettier version apprieciated.

### DIFF
--- a/templates/plugin/ceph.conf.erb
+++ b/templates/plugin/ceph.conf.erb
@@ -1,6 +1,14 @@
 <Plugin ceph>
-  LongRunAvgLatency <%= @longrunavglatency %>
-  ConvertSpecialMetricTypes <%= @convertspecialmetrictypes %>
+  <%- if @longrunavglatency -%>
+  LongRunAvgLatency true
+  <%- else -%>
+  LongRunAvgLatency false
+  <%- end -%>
+  <%- if @convertspecialmetrictypes -%>
+  ConvertSpecialMetricTypes true
+  <%- else -%>
+  ConvertSpecialMetricTypes false
+  <%- end -%>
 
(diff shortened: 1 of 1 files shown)


juniorsysadmin:
@vlitvin Current Puppet templating docs state that if the value isn’t a string, it will be automatically converted to a string using its to_s method . Out of interest, what Puppet and Ruby version are you using?

vlitvin:
@juniorsysadmin 
On client side:
  - ruby 2.0.0
  - puppet  3.8.1
On master:
  - ruby 2.1.5
  - puppet 3.7.2

igalic:
what about adding more validation to https://github.com/puppet-community/puppet-collectd/blob/master/manifests/plugin/ceph.pp ?

vlitvin:
@igalic validation is not a point. Not now however. Passing correct bool type causes incorrect content of config file on client side.

jyaworski:
Puppet4 is going to require strict typing. Also, does this account for when the values are 'false'?

```ruby
2.2.3 :006 > puts "I'm true" if 'true'
(irb):6: warning: string literal in condition
I'm true
 => nil 
2.2.3 :007 > puts "I'm true" if 'false'
(irb):7: warning: string literal in condition
I'm true
 => nil 
```

I feel adding validate_bool to the manifest is a better solution.

vlitvin:
Ok, strangely enough, I rolled back to  original ceph.conf.erb and everything seems fine.
I'm closing this pull request as I cannot reproduce behaviour.

PR #389, opened 2016-01-11
Allow easier creation of mysql monitoring through hiera

mnellemann:
Would something like this be possible for the mysql plugin ?
I've tested it locally w. my own manifests, but I don't know if it could cause problems elsewhere.

### DIFF
--- a/manifests/plugin/mysql.pp
+++ b/manifests/plugin/mysql.pp
@@ -4,6 +4,7 @@ class collectd::plugin::mysql (
   $ensure           = present,
   $manage_package   = $collectd::manage_package,
   $interval         = undef,
+  $databases        = { },
 ){
 
   if $::osfamily == 'Redhat' {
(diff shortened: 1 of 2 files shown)


manifests/plugin/mysql.pp
```diff
@@ -4,6 +4,7 @@
   $ensure           = present,
   $manage_package   = $collectd::manage_package,
   $interval         = undef,
+  $databases        = { },
```
danzilio: Can we set this to `false` or `undef` and wrap the call to `create_resources` in an `if` statement? We can use `validate_hash` to ensure that `$databases` is a hash.

igalic: create_resources takes all of that away from us
it does nothing on empty hash, and fails horribly if passed anything other than a (valid) hash.

danzilio: That's very considerate of `create_resources`! I withdraw my objection.

igalic:
would still be good to have this documented, @mnellemann!

jyaworski:
Hello:

Please rebase and/or squash.

PR #391, opened 2016-01-15
Make module.conf.erb support complex config

### LINKED ISSUE #390: collectd::plugin::python::module does not support complex config
The `collectd::plugin::python::module` defined type doesn't currently support config hashes with hash or array values.

For example, the declaration

```puppet
collectd::plugin::python::module { 'foo':
  module => 'foo',
  script_source => "puppet:///modules/mymodule/foo.py",
  config => {
    "key0" => ["e0", "e1"],
    "key1" => {"k0" => "v0", "k1" => "v1"},
  },
}
```

currently produces this configuration fragment:

```
  Import "foo"
  <Module "foo">
    key0 ["e0, "e1"]
    key1 {"k0" => "v0", "k1" => "v1"}
  </Module>
```

which is incorrect, and will make collectd fail with a configuration parse error at startup time.

Instead the following should be produced:

```
  Import "foo"
  <Module "foo">
    key0 "e0"
    key1 "e1"
    key1 "k0" "v0"
    key1 "k1" "v1"
  </Module>
```

If you agree I can probably work on a pull request.

elemoine:
This makes the module.conf.erb template support complex config, i.e. config
with values that are arrays or hashes.

Closes #390

### DIFF
--- a/spec/defines/collectd_plugin_python_module_spec.rb
+++ b/spec/defines/collectd_plugin_python_module_spec.rb
@@ -81,4 +81,21 @@ describe 'collectd::plugin::python::module', :type => :define do
                                              :path    => '/usr/share/collectd/python/foo.py',)
     end
   end
+
+  context 'complex config' do
+    let(:title) { 'foo' }
+    let :params do
+      {
+        :config => {'k1' => 'v1', 'k2' => ['v21', 'v22'], 'k3' => {'k31' => 'v31', 'k32' => 'v32'}},
+      }
+    end
+
+    it 'includes foo module configuration' do
+      should contain_concat__fragment('collectd_plugin_python_conf_foo').with(:content => /k1 v1/,)
+      should contain_concat__fragment('collectd_plugin_python_conf_foo').with(:content => /k2 v21/,)
+      should contain_concat__fragment('collectd_plugin_python_conf_foo').with(:content => /k2 v22/,)
+      should contain_concat__fragment('collectd_plugin_python_conf_foo').with(:content => /k3 k31 v31/,)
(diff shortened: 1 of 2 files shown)


elemoine:
The build failed on Travis, and I don't know why.

igalic:
because rubocop isn't being nice to its users

> Error: The `Style/TrailingComma` cop no longer exists. Please use `Style/TrailingCommaInLiteral` and/or `Style/TrailingCommaInArguments` instead.

elemoine:
Is there anything I can do about that?

robinbowes:
Yay for backwards compatibility and semantic versioning :/

igalic:
> Installing rubocop 0.36.0

so, technically it's totally semantically versioned. as long as you're < 1.0.0 all bets are off.

(which, btw, has been the reason why i started bumping all modules we release to 1.0.0 if they were 0.x.y)

robinbowes:
It has a version number that confirms the the semantic version standard. But, if new version is released that breaks backwards-compatibility, then it should be a major release. Hence, it is *not* semantically- versioned.

rnelson0:
But if you're under v1.x, there is no such thing as compatibility to break,
from a semver perspective. From http://semver.org/: `Major version zero
(0.y.z) is for initial development. Anything may change at any time. The
public API should not be considered stable.`

robinbowes:
Well, that sucks, doesn't it?! :)

igalic:
i honestly think we shouldn't derail this any further.
@mkrakowitzer has already created https://github.com/voxpupuli/modulesync_config/issues/17
and i honestly think now is probably as good a time as any to give in and do it.

PR #394, opened 2016-01-17
"Enable testmode switcher"

nibalizer:


igalic:
"would you mind changing your commit message"

I'd also appreciate an explanation what the motivation here is.
and finally: you do realize this will be overwritten next time someone does a modulesync, right? ;)

jyaworski:
I believe there is a facility in modulesync_config to add new gems. As for the class spec, I'm not sure.

juniorsysadmin:
Does this PR force Beaker testing to be done locally, even if say one might want to run the tests on docker, vms etc?

jyaworski:
@nibalizer is this something we should do across all of our modules?

nibalizer:
This was an experiment. The experiment has failed.

PR #393, opened 2016-01-15
Go back to collectd_version

jyaworski:
Someone forgot to rename everything....

PR #396, opened 2016-01-20
Fixing bug with rename collectd::params::package variable

mmogylenko:
Change-Id: I2dc6239228fc19da1bcfa426f75388529b9b3eae

In addition to https://github.com/voxpupuli/puppet-collectd/pull/393, last parameter that should be renamed in python plugin.


jyaworski:
Derp. Thanks!

PR #397, opened 2016-01-20
Fixes mongodb plugin params docs to be consistent

petems:
Rest of the document uses bullets rather than headings


igalic:
thanks @petems!

PR #403, opened 2016-02-03
Add support for puppetlabs-concat 2.x

brandonweeks:
* Bumps fixture versions of concat to 2.1.0
* Fixes two tests
* Removes $ensure from concat::fragment as this is no longer support in concat 2.x


dhoppe:
@brandonweeks Thank you very much for your contribution, but I think you have missed one ```ensure``` at [manifests/plugin/postgresql/database.pp](https://github.com/squarit/puppet-module-collectd/blob/master/manifests/plugin/postgresql/database.pp#L22)

brandonweeks:
Indeed I did! Thanks for catching that, I've updated the commit.

dhoppe:
:+1:

PR #409, opened 2016-02-16
Feature/curl json enhance

ortegaga:
This PR enhances the curl_json plugin with the following:
 - added Instance option to use within Key block
 - added Interval option
 - updated documentation with previous changes

It also ensures that package 'libyajl2' is installed in Debian (and derivatives)


ortegaga:
The Travist CI build tests have failed in plugins unrelated to the curl_json plugin

jyaworski:
Can you squash and/or rebase?

PR #410, opened 2016-02-17
Fixes issue with strict variables

petems:
`Undefined variable "collectd::manage_package"; class collectd has not been evaluated at /home/travis/build/voxpupuli/puppet-collectd/spec/fixtures/modules/collectd/manifests/plugin/mysql.pp:5 on node testing-worker-linux-docker-d88cd221-3376-linux-15.prod.travis-ci.org`


bastelfreak:
:+1:

PR #411, opened 2016-02-18
split out apache instance into define

ortegaga:
This PR adds a new defined type `collectd::plugin::apache::instance` that allows to automatize the configuration of apache monitoring in a central collectd server:

Consider the following use case:
Let's define the `profile::apache` class as: 
```
class profile::apache {
  # hiera stuff
  ...
  class { '::apache' :
    ...
  }

  @@collectd::plugin::apache::instance { $hostname :
    ...
  }
}
```
and the `profile::collectd::server` class as:
```
class profile::collectd::server {
  # hiera stuff
  ...
  Collectd::Plugin::Apache::Instance <<||>>
}
```
Whenever you apply the `profile::apache` class to a node, all the monitoring configuration of the apache instance will be 'automagically' generated on the central collectd server.

### DIFF
--- a/manifests/plugin/apache/instance.pp
+++ b/manifests/plugin/apache/instance.pp
@@ -1,0 +1,26 @@
+# https://collectd.org/wiki/index.php/Plugin:Apache
+define collectd::plugin::apache::instance (
+  $url,
+  $ensure = present,
+  $user = undef,
+  $password = undef,
+  $verifypeer = undef,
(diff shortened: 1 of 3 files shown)


jyaworski:
Can you give a bit more information about this?

ortegaga:
This is equivalent to PR #224, but allowing you to use exported resources of the defined type ``collectd::plugin::apache::instance``

PR #427, opened 2016-03-17
Quote python module blocks. Fixes #425

### LINKED ISSUE #425: Python module configs should be quoted
I ran into this while trying to deploy a module.

```shell
[root@jump1 ~]# cat /etc/collectd.d/python-config.conf 
# Generated by Puppet
<Plugin "python">
  ModulePath "/usr/share/collectd/python"
  LogTraces false
  Interactive false
  Import "puppet_reports"

  <Module "puppet_reports">
    ReportsDir /var/lib/puppet/reports
  </Module>
</Plugin>
[root@jump1 ~]# service collectd stop
Stopping collectd:                                         [  OK  ]
[root@jump1 ~]# service collectd start
Starting collectd: Parse error in file `/etc/collectd.d/python-config.conf', line 136 near `/': syntax error, unexpected SLASH
yyparse returned error #1
configfile: Cannot read file `/etc/collectd.d/python-config.conf'.
Unable to read config file /etc/collectd.conf.
Error: Reading the config file failed!
Read the syslog for details.
not starting due to configuration error
```

YAML used for configuration via hiera:

```yaml
collectd::plugin::python::modules:
  'puppet_reports':
    config:
      ReportsDir: '/var/lib/puppet/reports'
    script_source: 'puppet:///modules/profiles/collectd/python/puppet_reports.py
```

jyaworski:


jyaworski:
I need to actually think about how to do this. Sometimes there strings already wrapped in ", sometimes not. I could look at value[-1] and such, but that leaves me wondering if collectd actually interprets true/false as booleans.

igalic:
you could use `Shellwords#split`
that should give you the naked tokens

jyaworski:
@igalic doing that gives me, for example:

```
 k1 "["v1"]"
```

jyaworski:
Referencing [0] works:

```
k1 "v1"
```

Now I need to test for booleans.

jyaworski:
Got it. Thanks @igalic.

PR #434, opened 2016-03-30
Plugin options do not use equals signs when assigning values

cqwense:


jyaworski:
Is this collectd version dependent?

cqwense:
I do not believe so, I'm going off the suggestion of the team in my org who owns our collectd configuration's assertion, and the fact that I cannot find any examples ( of any collectd version ) that uses ='s in the plugin options config.

PR #436, opened 2016-04-05
add optional header param for collectd::plugin::curl_json

TheMeier:
* add param $header = undef to collectd::plugin::curl_json
* add new param to collectd_plugin_curl_json_spec.rb
* add rspec test for collectd::plugin::curl_json


jyaworski:
Is this the HTTP headers for the curl command? If so, can you make it a hash of hashes?

PR #451, opened 2016-04-16
ReportsReserved only exists in the df plugin before 5.5. Fixes #435

### LINKED ISSUE #435: Plugin `df' did not register for value `ReportReserved'.
According to the documentation, there is no parameter ReportReserved defined for the df plugin
https://collectd.org/documentation/manpages/collectd.conf.5.shtml#plugin_df

The corresponding template does however add it, and as a result I get an error log as seen in the title of this issue. This is at least the case for collectd 5.5

It should be possible to make this value optional (e.g. by assigning a default **undef** value to it)

jyaworski:
@xaniasd how does this look?

### DIFF
--- a/templates/plugin/df.conf.erb
+++ b/templates/plugin/df.conf.erb
@@ -8,7 +8,9 @@
   IgnoreSelected <%= @ignoreselected %>
   ReportByDevice <%= @reportbydevice %>
   ReportInodes <%= @reportinodes %>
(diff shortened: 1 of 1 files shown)


nibalizer:
:+1: after the fact

xaniasd:
@jyaworski awesome! thanks a lot

PR #454, opened 2016-04-18
Set collectd_version fact to empty string if collectd not installed

### LINKED ISSUE #421: write_graphite plugin config not setting protocol on first run
See: https://github.com/collectd/collectd/issues/866

In write_graphite config, the protocol value is correctly written if you have collectd 5.4 installed. If you're running a later version, then it's not set. See https://github.com/voxpupuli/puppet-collectd/blob/master/templates/plugin/write_graphite/carbon.conf.erb#L16

johanek:
Fixes #421 - ```collectd_real_version``` in ```init.pp``` not set correctly as the fact returns ```nil```, and ```pick``` will accept that as a value.

### DIFF
--- a/lib/facter/collectd_version.rb
+++ b/lib/facter/collectd_version.rb
@@ -11,6 +11,8 @@ Facter.add(:collectd_version) do
     if Facter::Util::Resolution.which('collectd')
       collectd_help = Facter::Util::Resolution.exec('collectd -h')
       %r{^collectd ([\w\.]+), http://collectd\.org/}.match(collectd_help)[1]
+    else
+      ''
     end
   end
 end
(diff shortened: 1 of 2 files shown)


bastelfreak:
Thanks for the contribution!

jyaworski:
@bastelfreak @johanek does pick ignore ''?

johanek:
Yes, it needs to be an emtpy string or undefined rather than nil 

https://github.com/puppetlabs/puppetlabs-stdlib#pick

> From a list of values, returns the first value that is not undefined or an empty string. Takes any number of arguments, and raises an error if all values are undefined or empty.

Issue #418, opened 2016-03-03
collectd::plugin::python::module does not support nested configuration


aequitas:
For example: https://github.com/NYTimes/collectd-rabbitmq#example-configuration

jyaworski:
Setting multiple modules introduces multiple concat fragments. What are you seeing when you encounter this?

jyaworski:
Ping @aequitas

aequitas:
Sorry missed your question. I will come back with an example on Monday.

aequitas:
The problem I have is the rabbitmq plugin can have nested configuration blocks inside its configuration, for example with the `<Ignore>` block below:

```
LoadPlugin python
<Plugin python>
  <Module rabbitmq>
    Host "localhost"
    <Ignore "queue">
      Regex "amq-gen-.*"
    </Ignore>
  </Module>
</Plugin>
```

The Python plugin only support key/value pairs. I have a workaround which works atm:

```
        collectd::plugin::python::module { 'rabbitmq':
            module        => 'rabbitmq',
            script_source => 'puppet:///modules/collectd/rabbitmq.py',
            config        => {
                '<Ignore "queue">' => "\nRegex \"[0-9a-z]{32}\"\nRegex \"celeryev\\..*\"\n</Ignore>",
            },
        }
```

Maybe having just a freeform config append variable would suffice. I think implementing nested support would make it a lot to complicated for simple exceptional use cases.

jyaworski:
@aequitas just curious, why use the python plugin for rabbitmq vs curl_json?

I like the idea of a freeform config.

aequitas:
Tbh, it was the first hit in Google and it fit our usecase  :)

jyaworski:
I'll see if I can get to this sometime soon. I also have to tackle #456.

I would check out curl_json. There's actually an example [in our docs](https://github.com/voxpupuli/puppet-collectd#class-collectdplugincurl_json). I've found that the perl and python plugins should be avoided when possible due to their overhead.

aequitas:
thanks, will be sure to check it out!

microblag:
I've been struggling with this same issue. Here's the hackyish bit of erb I've thrown together to try and tackle it. I couldn't come up with a good way to represent the config data structure so there's a couple of quirks. It's got no error checking or validation also I'm no Ruby guy so I may be horribly miss-using Proc. So I'm providing it here in case it's a useful for someone having a proper go at adding this functionality.

[...]

PR #468, opened 2016-04-20
Rabbitmq

bilsch:
Adding support for rabbitmq monitoring.

This is mostly complete / ready. I still need to look at the TypesDB line - I think that is just a config thing on my part.

The spec is mostly complete though at present it does not check a provider other than pip. I don't have other providers ( eg, rpm deb / apt etc ) available for testing. Can be added to the spec if desired.

### DIFF
--- a/manifests/plugin/rabbitmq.pp
+++ b/manifests/plugin/rabbitmq.pp
@@ -1,0 +1,75 @@
+# rabbitmq plugin
+# https://pypi.python.org/pypi/collectd-rabbitmq
+#
+# == Class collectd::plugin::rabbitmq
+#
+#  Configures rabbitmq metrics collection. Optionally installs the plugin
+#  Note, it is up to you to support package installation and sources
+#
+# === Parameters:
+# [*ensure*]
+#   String
+#   Passed to package and collectd::plugin resources ( both )
+#   Default: present
+#
+# [*interval*]
+#   Integer
+#   Interval setting for the plugin
+#   Default: undef
+#
+# [*manage_package*]
+#   Boolean
+#   Toggles installation of plugin. Please reference https://collectd-rabbitmq.readthedocs.org/en/latest/installation.html
+#   Default: undef
+#
+# [*package_provider*]
+#   String
+#   Passed to package resource
+#   Default: pip
+#
+# [*config*]
+#   Hash
+#   Contains key/value passed to the python module to configure the plugin
+#   Default: {
+#    'Username' => '"guest"',
+#    'Password' => '"guest_pass"',
+#    'Scheme'   => '"http"',
+#    'Port'     => '"15672"',
+#    'Host'     => "\"${::fqdn}\"",
+#    'Realm'    => '"RabbitMQ Management"',
+#   }
+#
+class collectd::plugin::rabbitmq (
+  $config           = {
+    'Username' => '"guest"',
+    'Password' => '"guest"',
+    'Scheme'   => '"http"',
+    'Port'     => '"15672"',
+    'Host'     => "\"${::fqdn}\"",
+    'Realm'    => '"RabbitMQ Management"',
+  },
+  $ensure           = 'present',
+  $interval         = undef,
+  $manage_package   = undef,
+  $package_name     = 'collectd-rabbitmq',
+  $package_provider = 'pip',
+) {
+  include ::collectd
+
+  validate_string($ensure)
+  validate_hash($config)
+
+  $_manage_package = pick($manage_package, $::collectd::manage_package)
+
+  if $_manage_package {
+    package { $package_name:
+      ensure   => $ensure,
+      provider => $package_provider,
+    }
+  }
+  collectd::typesdb { '/usr/share/collect-rabbitmq/types.db.custom': }
+  collectd::plugin::python::module { 'collectd_rabbitmq.collectd_plugin':
+    ensure => $ensure,
+    config => $config,
+  }
+}
--- a/README.md
+++ b/README.md
@@ -119,6 +119,7 @@ documentation for each plugin for configurable attributes.
 * `protocols` (see [collectd::plugin:protocols](#class-collectdpluginprotocols) below)
 * `python` (see [collectd::plugin::python](#class-collectdpluginpython) below)
 * `redis` (see [collectd::plugin::redis](#class-collectdpluginredis) below)
+* `rabbitmq` (see [collectd-rabbitmq](https://pypi.python.org/pypi/collectd-rabbitmq) and [below](#class-collectdpluginrabbitmq) for implementation notes
 * `rrdcached` (see [collectd::plugin::rrdcached](#class-collectdpluginrrdcached) below)
 * `rrdtool` (see [collectd::plugin::rrdtool](#class-collectdpluginrrdtool) below)
 * `sensors` (see [collectd::plugin::sensors](#class-collectdpluginsensors) below)
@@ -1226,6 +1227,23 @@ class { 'collectd::plugin::redis':
       'timeout'  => 3000,
     }
   }
+}
+```
+
+####Class: `collectd::plugin::rabbitmq`
+
+Please note the rabbitmq plugin provides a [types.db.custom](https://github.com/NYTimes/collectd-rabbitmq/blob/master/config/types.db.custom). You will need to add this to [collectd::typesdb](https://github.com/voxpupuli/puppet-collectd/blob/master/manifests/init.pp#L28) via hiera or in a manifest. Failure to set the types.db.custom content will result in *no* metrics from the rabbitmq plugin.
+
+```puppet
+class { '::collectd::plugin::rabbitmq':
+  config           => {
+    'Username' => '"admin"',
+    'Password' => "${admin_pass}",
+    'Scheme'   => '"https"',
+    'Port'     => '"15671"',
+    'Host'     => "${::fqdn}",
+    'Realm'    => '"RabbitMQ Management"',
+  },
 }
 ```
 
--- a/spec/classes/collectd_plugin_rabbitmq_spec.rb
+++ b/spec/classes/collectd_plugin_rabbitmq_spec.rb
@@ -1,0 +1,129 @@
+require 'spec_helper'
+
+describe 'collectd::plugin::rabbitmq', type: :class do
+  let :facts do
+    {
+      osfamily: 'RedHat',
+      collectd_version: '5.5.1'
+    }
+  end
+
+  context 'package ensure' do
+    context ':ensure => present' do
+      let(:node) { 'testhost.example.com' }
+
+      it 'Load collectd_rabbitmq in python-config' do
+        should contain_concat_fragment('collectd_plugin_python_conf_collectd_rabbitmq.collectd_plugin').with_content(/Module "collectd_rabbitmq.collectd_plugin"/)
+      end
+
+      it 'import collectd_rabbitmq.collectd_plugin in python-config' do
+        should contain_concat_fragment('collectd_plugin_python_conf_collectd_rabbitmq.collectd_plugin').with_content(/Import "collectd_rabbitmq.collectd_plugin"/)
+      end
+
+      it 'default to Username guest in python-config' do
+        should contain_concat_fragment('collectd_plugin_python_conf_collectd_rabbitmq.collectd_plugin').with_content(/Username "guest"/)
+      end
+
+      it 'default to Password guest in python-config' do
+        should contain_concat_fragment('collectd_plugin_python_conf_collectd_rabbitmq.collectd_plugin').with_content(/Password "guest"/)
+      end
+
+      it 'default to Port 15672 in python-config' do
+        should contain_concat_fragment('collectd_plugin_python_conf_collectd_rabbitmq.collectd_plugin').with_content(/Port "15672"/)
+      end
+
+      it 'default to Scheme http in python-config' do
+        should contain_concat_fragment('collectd_plugin_python_conf_collectd_rabbitmq.collectd_plugin').with_content(/Scheme "http"/)
+      end
+
+      it 'Host should be set to $::fqdn python-config' do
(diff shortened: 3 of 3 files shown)


jyaworski:
There's been some talk about this in #418.

Can you squash?

bilsch:
already squashed sorry about that!

bilsch:
I had some issues when I left the `include collectd` in our setup - I got a duplicate def. I'll add an ensure_resource to that which should cover things. May just be a bug in how I wrote the profile on our end.

I'll look at this more tomorrow morning

manifests/plugin/rabbitmq.pp
```diff
@@ -0,0 +1,56 @@
...
+  $config           = {
+    'user'     => 'guest',
+    'password' => 'guest',
+    'host'     => $::fqdn,
+    'port'     => 15672,
+    'scheme'   => 'http',
+  },
+  $ensure           = 'present',
+  $interval         = $::collectd::params::interval,
+  $manage_package   = false,
```
jyaworski: We generally default to undef here, and do a pick($manage_package, $::collectd::manage_package).

manifests/plugin/rabbitmq.pp
```diff
@@ -0,0 +1,56 @@
...
+class collectd::plugin::rabbitmq (
+  $config           = {
+    'user'     => 'guest',
+    'password' => 'guest',
+    'host'     => $::fqdn,
+    'port'     => 15672,
+    'scheme'   => 'http',
+  },
+  $ensure           = 'present',
+  $interval         = $::collectd::params::interval,
```
jyaworski: This will fail strict_variables, likely. Set it to undef, please.

manifests/plugin/rabbitmq.pp
```diff
@@ -0,0 +1,56 @@
...
+  $manage_package   = false,
+  $package_provider = 'pip',
+) {
+
+  if $manage_package {
```
jyaworski: Does this exist on Debian-ish as well? The rest of the plugins check for RedHat as an osfamily.

templates/plugin/rabbitmq.conf.erb
```diff
@@ -0,0 +1,19 @@
+# TODO need to find a better way to handle the types.db config
+TypesDB "/usr/share/collectd/types.db"
+TypesDB "/usr/share/collect-rabbitmq/types.db.custom"
+
+LoadPlugin python
+<Plugin python>
```
jyaworski: This seems remarkably similar to the way that genericjmx is written. Can you refactor around the way the python plugin is written?

jyaworski:
Look at some other plugins. I would include collectd and collectd::plugin::python in the manifest (similar to the rabbitmq plugin). This is how we get around testing with strict_variables.

bilsch:
@jyaworski thanks for the feeback I'll get to work on the refactor.

bilsch:
Trying to run the [acceptance tests](https://github.com/voxpupuli/puppet-collectd/blob/master/CONTRIBUTING.md#integration-tests) locally and running into some trouble. Seems the base box / config does not have access to install collectd from upstream?

```
Beaker::Hypervisor, found some vagrant boxes to create
Bringing machine 'centos-66-x64' up with 'virtualbox' provider...
==> centos-66-x64: Importing base box 'puppetlabs/centos-6.6-64-nocm'...
Progress: 10%Progress: 20%Progress: 30%Progress: 40%Progress: 50%Progress: 60%Progress: 70%Progress: 80%Progress: 90%==> centos-66-x64: Matching MAC address for NAT networking...
==> centos-66-x64: Checking if box 'puppetlabs/centos-6.6-64-nocm' is up to date...
==> centos-66-x64: Setting the name of the VM: centos-66-x64yml_centos-66-x64_1461246999316_11567
==> centos-66-x64: Pruning invalid NFS exports. Administrator privileges will be required...
==> centos-66-x64: Clearing any previously set network interfaces...
==> centos-66-x64: Preparing network interfaces based on configuration...
    centos-66-x64: Adapter 1: nat
    centos-66-x64: Adapter 2: hostonly
==> centos-66-x64: Forwarding ports...
    centos-66-x64: 22 (guest) => 2222 (host) (adapter 1)
==> centos-66-x64: Running 'pre-boot' VM customizations...
==> centos-66-x64: Booting VM...
==> centos-66-x64: Waiting for machine to boot. This may take a few minutes...
    centos-66-x64: SSH address: 127.0.0.1:2222
    centos-66-x64: SSH username: vagrant
    centos-66-x64: SSH auth method: private key
==> centos-66-x64: Machine booted and ready!
==> centos-66-x64: Checking for guest additions in VM...
==> centos-66-x64: Setting hostname...
==> centos-66-x64: Configuring and enabling network interfaces...
==> centos-66-x64: Mounting shared folders...
    centos-66-x64: /vagrant => /Users/bschwanitz/gits/cmm/puppet-collectd/.vagrant/beaker_vagrant_files/centos-66-x64.yml
Warning: ssh connection to centos-66-x64 has been terminated
Warning: ssh connection to centos-66-x64 has been terminated
Disabling updates.puppetlabs.com by modifying hosts file to resolve updates to 127.0.0.1 on centos-66-x64
localhost $ scp /Users/bschwanitz/gits/cmm/puppet-collectd centos-66-x64:/etc/puppet/modules {:ignore => [".bundle", ".git", ".idea", ".vagrant", ".vendor", "vendor", "acceptance", "bundle", "spec", "tests", "log", ".", ".."]}

collectd class
  default parameters
localhost $ scp /var/folders/sw/1rq791qd3kb5hvg1xcxknzrx7h72wx/T/beaker20160421-76717-1y42mw2 centos-66-x64:/tmp/apply_manifest.pp.CLCL0n {:ignore => }
    should work idempotently with no errors (FAILED - 1)
    Package "collectd"
      should be installed (FAILED - 2)
    Service "collectd"
      should be running (FAILED - 3)
Warning: ssh connection to centos-66-x64 has been terminated
Destroying vagrant boxes
==> centos-66-x64: Forcing shutdown of VM...
==> centos-66-x64: Destroying VM and associated drives...

Failures:

  1) collectd class default parameters should work idempotently with no errors
     Failure/Error: apply_manifest(pp, :catch_failures => true)
     Beaker::Host::CommandFailure:
       Host 'centos-66-x64' exited with 4 running:
        puppet apply --verbose --detailed-exitcodes /tmp/apply_manifest.pp.CLCL0n
       Last 10 lines of output were:
       	Notice: /Stage[main]/Collectd::Config/File_line[include_conf_d_dot_conf]: Dependency Package[collectd] has failures: true
       	Warning: /Stage[main]/Collectd::Config/File_line[include_conf_d_dot_conf]: Skipping because of failed dependencies
       	Notice: /Stage[main]/Collectd::Config/File_line[include_conf_d]: Dependency Package[collectd] has failures: true
       	Warning: /Stage[main]/Collectd::Config/File_line[include_conf_d]: Skipping because of failed dependencies
       	Notice: /Stage[main]/Collectd::Config/File[collectd.d]: Dependency Package[collectd] has failures: true
       	Warning: /Stage[main]/Collectd::Config/File[collectd.d]: Skipping because of failed dependencies
       	Notice: /Stage[main]/Collectd::Service/Service[collectd]: Dependency Package[collectd] has failures: true
       	Warning: /Stage[main]/Collectd::Service/Service[collectd]: Skipping because of failed dependencies
       	Info: Creating state file /var/lib/puppet/state/state.yaml
       	Notice: Finished catalog run in 0.37 seconds

     # ./spec/acceptance/class_spec.rb:12:in `block (3 levels) in <top (required)>'

  2) collectd class default parameters Package "collectd" should be installed
     Failure/Error: it { should be_installed }
       expected Package "collectd" to be installed

     # ./spec/acceptance/class_spec.rb:17:in `block (4 levels) in <top (required)>'

  3) collectd class default parameters Service "collectd" should be running
     Failure/Error: it { should be_running }
       expected Service "collectd" to be running

     # ./spec/acceptance/class_spec.rb:21:in `block (4 levels) in <top (required)>'

Finished in 17.8 seconds (files took 55.29 seconds to load)
3 examples, 3 failures

Failed examples:

rspec ./spec/acceptance/class_spec.rb:6 # collectd class default parameters should work idempotently with no errors
rspec ./spec/acceptance/class_spec.rb:17 # collectd class default parameters Package "collectd" should be installed
rspec ./spec/acceptance/class_spec.rb:21 # collectd class default parameters Service "collectd" should be running
```

So far I've only tried `centos-66-x64` and `ubuntu-server-1404-x64`. Is there something not documented or that I missed to run these locally?

jyaworski:
@bilsch we generally only do the rspec portion of the tests, for the purposes of this PR. However, it looks like you've found out our acceptance tests don't work.

@bastelfreak @nibalizer

bilsch:
So, after all of my tinkering on this I'm re-considering the need for this PR. Looking for input from you guys before I close it out.

My profile:

```
(...)
  require profile::base::metrics
(...)
  package { 'collectd-rabbitmq':
    provider => 'pip',
  } ->
  collectd::plugin::python::module {'collectd_rabbitmq.collectd_plugin':
    config => {
      'Username' => '"admin"',
      'Password' => "\"${admin_pass}\"",
      'Scheme'   => '"https"',
      'Port'     => '"15671"',
      'Host'     => "\"${::fqdn}\"",
      'Realm'    => '"RabbitMQ Management"',
    },
  }
```

I then added an array of typesdb files and all is working. We have collectd inclusion happening in profile::base::metrics, fixed the resource ordering issue by requiring that to load before the python::module plugin.

In short, it was not all that hard to configure this without hacking on the module itself. Thoughts @jyaworski ?

jyaworski:
I honestly think this is very useful to have as its own plugin, like genericjmx is. It's a wrapper, but it's used enough to be included directly.

@aequitas would also appreciate it.

bilsch:
specs are passing locally but I have not yet tested this version. Is this closer to what you were thinking @jyaworski ? I basically took what I did in our profile and put it in the rabbitmq class. I have a little bit of cleanup to do and still need to think about how to address the types.db.custom provided by the python module so this is not quite yet ready

.gitignore
```diff
@@ -8,4 +8,5 @@ coverage/
 log/
 .idea/
 *.iml
-.*.sw
+.*.sw*
+.DS_Store
```
jyaworski: This will be overwritten. Please open a PR against https://github.com/voxpupuli/modulesync_config/blob/master/moduleroot/.gitignore

jyaworski:
@bilsch looks good so far. Looking forward to the finished product.

P.S. I added [WIP] to the PR title so nobody merges it. Please remove it when you're ready.

bilsch:
Ok, I've tested the module and everything seems in order. I'm not sure how to resolve the issue of the types.db.custom provided by the python module.  I could do something like [typesdb](https://github.com/voxpupuli/puppet-collectd/blob/master/examples/typesdb.pp#L9) but this seems to wipe out the contents. Any suggestions there @jyaworski 

The only other option I know of so far is to force people to add it to `collectd::typesdb` array but I don't see a way to do that within the manifest

manifests/plugin/rabbitmq.pp
```diff
@@ -0,0 +1,74 @@
...
+    'Username' => '"guest"',
+    'Password' => '"guest"',
+    'Scheme'   => '"http"',
+    'Port'     => '"15672"',
+    'Host'     => "\"${::fqdn}\"",
+    'Realm'    => '"RabbitMQ Management"',
+  },
+  $ensure           = 'present',
+  $interval         = undef,
+  $manage_package   = undef,
```
jyaworski: Can you add a package_name parameter here and set it to collectd-rabbitmq by default?

manifests/plugin/rabbitmq.pp
```diff
@@ -0,0 +1,74 @@
...
+  include ::collectd
+
+  validate_string($ensure)
+  validate_hash($config)
+
+  $_manage_package = pick($manage_package, $::collectd::manage_package)
+
+  if $_manage_package {
+    package { 'collectd-rabbitmq':
+      ensure   => $ensure,
```
jyaworski: And then set name => $package_name here.

jyaworski:
collectd::typesdb is a defined type, so you can add the array of defined types in the plugin manifest.

bilsch:
I was referring to the [init.pp](https://github.com/voxpupuli/puppet-collectd/blob/master/manifests/init.pp#L28) class setting - basically add a value to that [array](https://github.com/voxpupuli/puppet-collectd/blob/master/manifests/params.pp#L16)

did the package_name and added specs for that as well. Thoughts on the best way to handle the typesdb stuff @jyaworski ? My current solution is to just use hiera and set to the rpm and collectd-rabbitmq typesdb files and all works well

jyaworski:
Are you looking to add types, and not another typesdb? @bastelfreak @alvin.huang

bilsch:
The plugin drops a [types.db.custom](https://github.com/NYTimes/collectd-rabbitmq/blob/master/config/types.db.custom) which needs to be loaded in addition to the collectd default types.db. The other option which I kinda don't want to do is to add those types in via this rabbitmq class - too delicate for my tastes

jyaworski:
@bilsch really? Passing the array of types it declares should be pretty straightforward.

bilsch:
> @bilsch really? Passing the array of types it declares should be pretty straightforward.

My main thought behind being delicate is what happens when the types change. We now have config data for a 3rd party in the puppet module. When the package provides it as a file it would be better to just use that ;)

I'll add the types to the change and that should resolve the last item. Thanks again for your input

jyaworski:
Ah, OK. You're concerned about upstream changing the typesdb. OK, that's completely reasonable.

bilsch:
The last push failed to validate due to gem install conflicts. Seems to be an issue with ruby <= 2.1.

I updated the docs to reflect the types.db requirement. I think short of putting those types into the manifest ( which I'm personally not in favor of ) this Pr should be ready for final review / merge.

jyaworski:
@bilsch the gem requirement problem is something @bastelfreak and I are working on across all of our modules. The 'listen' gem updated to 3.1, which made the minimum Ruby requirement 2.2. If your tests pass on 2.2 (which they did) I consider this ready.

PR #469, opened 2016-04-20
TimerPercentile can be specified multiple times

kimor79:
As is, this is a backwards incompatible change. Is there a way either in the manifest or the template to check if the value is an array or scalar first?


manifests/plugin/statsd.pp
```diff
@@ -8,7 +8,7 @@
   $deletegauges    = undef,
   $deletesets      = undef,
   $interval        = undef,
-  $timerpercentile = undef,
+  $timerpercentile = [],
```
jyaworski: Setting a parameter to [] in class parameters confuses hiera. Please use pick() inside the manifest and default to undef here.

PR #470, opened 2016-04-21
Update based on voxpupuli/modulesync_config

dhoppe:
Please merge.


lib/puppet/parser/functions/collectd_convert_processmatch.rb
```diff
@@ -1,4 +1,4 @@
-Puppet::Parser::Functions.newfunction(:collectd_convert_processmatch, :type => :rvalue, :arity => 1, :doc => <<-ENDDOC
+Puppet::Parser::Functions.newfunction(:collectd_convert_processmatch, type: :rvalue, arity: 1, doc: <<-ENDDOC
```
jyaworski: It looks as if this is the only change that actually breaks 1.8.7. I guess I'll have to modify locally. :)

Gemfile
```diff
@@ -58,7 +58,7 @@ else
 gem 'facter', :require => false, :groups => [:test]
 end
 
-ENV['PUPPET_VERSION'].nil? ? puppetversion = '3.8.5' : puppetversion = ENV['PUPPET_VERSION'].to_s
+ENV['PUPPET_VERSION'].nil? ? puppetversion = '~> 3.0' : puppetversion = ENV['PUPPET_VERSION'].to_s
```
jyaworski: This didn't pick up the 4.0 change? The same with .travis.yml and the new matrix.

dhoppe: My fault. I forgot the rebase between upstream and master. :(

PR #472, opened 2016-04-22
add manage_service param

archii:
Added a parameter to toggle automatic starting of the collectd service.  
The default behavior is the same as before, which is to automatically start the service.


jyaworski:
Can you update the README and add a spec test?

jyaworski:
@archii it looks like the spec test is failing. Can you take a look?

archii:
The tests are failing on trying to install the listen gem v 3.1.1 on Ruby version <2.2.  I see that this is the case in master as well.  Not sure what changed.  Having a look...

jyaworski:
@archii this is good. Take a look at #475 where I'm fixing that problem. It's not your code, so I'm merging.

Issue #456, opened 2016-04-18
Python plugin regression


jyaworski:
I believe #352 introduced this problem:

```puppet
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Failed to parse template collectd/plugin/python/module.conf.erb:
  Filepath: org/jruby/RubyModule.java
  Line: 2733
  Detail: uninitialized constant Puppet::Parser::TemplateWrapper::Shellwords
 at /etc/puppet/environments/production/modules/collectd/manifests/plugin/python/module.pp:37 on node puppet1.domain.com
Warning: Not using cache on failed catalog
Error: Could not retrieve catalog; skipping run
```

bilsch:
I keep running in to this with the rabbitmq change I put up. Going to take a stab at tracking this down

jyaworski:
@bilsch you're awesome. Thanks.

bilsch:
So, turns out my rabbitmq module was not as ready as I thought and oops

```
  1) collectd::plugin::rabbitmq package ensure :ensure => present Realm set to "RabbitMQ Management"
     Failure/Error: should contain_concat_fragment('collectd_plugin_python_conf_collectd_rabbitmq.collectd_plugin').with_content(/Realm "RabbitMQ Management"/)

       expected that the catalogue would contain Concat_fragment[collectd_plugin_python_conf_collectd_rabbitmq.collectd_plugin] with content set to /Realm "RabbitMQ Management"/ but it is set to "  Import \"collectd_rabbitmq.collectd_plugin\"\n\n  <Module \"collectd_rabbitmq.collectd_plugin\">\n    Host \"testhost.example.com\"\n    Password \"guest\"\n    Port \"15672\"\n    Realm \"RabbitMQ\" \"Management\"\n    Scheme \"http\"\n    Username \"guest\"\n  </Module>\n"
       Diff:
       @@ -1,2 +1,11 @@
       -(?-mix:Realm "RabbitMQ Management")
       +  Import "collectd_rabbitmq.collectd_plugin"
       +
       +  <Module "collectd_rabbitmq.collectd_plugin">
       +    Host "testhost.example.com"
       +    Password "guest"
       +    Port "15672"
       +    Realm "RabbitMQ" "Management"
       +    Scheme "http"
       +    Username "guest"
       +  </Module>
```

Looks like an issue in the logic introduced with https://github.com/voxpupuli/puppet-collectd/pull/427

Wonder if its just an order thing?

PR #478, opened 2016-04-25
[WIP] fixes for python module shellwords issues

bilsch:
I would like to know if this patch set addresses #418 @aequitas would you mind trying my fork of the module?

This patch set fixes a few things. The rabbitmq change ( #468 ) was mostly there but I did not notice an issue with the typesdb change. This should now be fixed.

This should address #456 but would like some additional validation.

This patch is addressing a few things - should I break it up or no?

### DIFF
--- a/templates/plugin/python/module.conf.erb
+++ b/templates/plugin/python/module.conf.erb
@@ -11,7 +11,11 @@
     <%= key %> <%= k %> <% if !!v == v %><%= v %><% else %>"<%= Shellwords.split(v).join('" "') %>"<% end %>
   <%- end -%>
   <%- else -%>
-    <%= key %> <% if !!value == value %><%= value %><% else %>"<%= Shellwords.split(value).join('" "') %>"<% end %>
+  <%- if not !!value == value and value.include?('"') -%>
+    <%= key %> "<%= Shellwords.split(value).join('" "') %>"
+  <%- else -%>
+    <%= key %> <% if !!value == value %><%= value %><% else %>"<%= value %>"<% end %>
+  <%- end -%>
   <%- end -%>
   <%- end -%>
   </Module>
--- a/README.md
+++ b/README.md
@@ -1241,17 +1241,25 @@ class { 'collectd::plugin::redis':
 
 ####Class: `collectd::plugin::rabbitmq`
 
-Please note the rabbitmq plugin provides a [types.db.custom](https://github.com/NYTimes/collectd-rabbitmq/blob/master/config/types.db.custom). You will need to add this to [collectd::typesdb](https://github.com/voxpupuli/puppet-collectd/blob/master/manifests/init.pp#L28) via hiera or in a manifest. Failure to set the types.db.custom content will result in *no* metrics from the rabbitmq plugin.
+Please note the rabbitmq plugin provides a [types.db.custom](https://github.com/NYTimes/collectd-rabbitmq/blob/master/config/types.db.custom). You will need to add this to [collectd::config::typesdb](https://github.com/voxpupuli/puppet-collectd/blob/master/manifests/config.pp#L18) via hiera or in a manifest. Failure to set the types.db.custom content will result in *no* metrics from the rabbitmq plugin.
+
+set typesdb to include the collectd-rabbitmq types.db.custom
+
+```yaml
+collectd::config::typesdb:
+  - /usr/share/collectd/types.db
+  - /usr/share/collect-rabbitmq/types.db.custom
+```
 
 ```puppet
 class { '::collectd::plugin::rabbitmq':
   config           => {
-    'Username' => '"admin"',
-    'Password' => "${admin_pass}",
-    'Scheme'   => '"https"',
-    'Port'     => '"15671"',
-    'Host'     => "${::fqdn}",
-    'Realm'    => '"RabbitMQ Management"',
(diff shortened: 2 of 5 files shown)


templates/plugin/python/module.conf.erb
```diff
@@ -11,7 +11,11 @@
     <%= key %> <%= k %> <% if !!v == v %><%= v %><% else %>"<%= Shellwords.split(v).join('" "') %>"<% end %>
   <%- end -%>
   <%- else -%>
-    <%= key %> <% if !!value == value %><%= value %><% else %>"<%= Shellwords.split(value).join('" "') %>"<% end %>
+  <%- if not !!value == value and value.include?('"') -%>
```
jyaworski: How does this address the shellwords issue?

bilsch: If I set the test case back to `    <%= key %> <% if !!value == value %><%= value %><% else %>"<%= Shellwords.split(value).join('" "') %>"<% end %>
`

I get a funky output / test failure.

```
Failures:

  1) collectd::plugin::rabbitmq package ensure :ensure => present Realm set to "RabbitMQ Management"
     Failure/Error: should contain_concat_fragment('collectd_plugin_python_conf_collectd_rabbitmq.collectd_plugin').with_content(/Realm "RabbitMQ Management"/)

       expected that the catalogue would contain Concat_fragment[collectd_plugin_python_conf_collectd_rabbitmq.collectd_plugin] with content set to /Realm "RabbitMQ Management"/ but it is set to "  Import \"collectd_rabbitmq.collectd_plugin\"\n\n  <Module \"collectd_rabbitmq.collectd_plugin\">\n    Host \"testhost.example.com\"\n    Password \"guest\"\n    Port \"15672\"\n    Realm \"RabbitMQ\" \"Management\"\n    Scheme \"http\"\n    Username \"guest\"\n  </Module>\n"
       Diff:
       @@ -1,2 +1,11 @@
       -(?-mix:Realm "RabbitMQ Management")
       +  Import "collectd_rabbitmq.collectd_plugin"
       +
       +  <Module "collectd_rabbitmq.collectd_plugin">
       +    Host "testhost.example.com"
       +    Password "guest"
       +    Port "15672"
       +    Realm "RabbitMQ" "Management"
       +    Scheme "http"
       +    Username "guest"
       +  </Module>

     # ./spec/classes/collectd_plugin_rabbitmq_spec.rb:56:in `block (4 levels) in <top (required)>'
```

At least in my use-case previously I was getting the ShellWord exception on render. It also addresses the spam issue:

Force the `!!value == value` to fail by changing to `  <%- if not value == value and value.include?('XXXXXX') -%>`

```
Failures:

  1) collectd::plugin::python::module spam module includes spam module configuration
     Failure/Error:
       should contain_concat__fragment('collectd_plugin_python_conf_spam')
         .with(content: /spam "wonderful" "lovely"/,)

       expected that the catalogue would contain Concat::Fragment[collectd_plugin_python_conf_spam] with content set to /spam "wonderful" "lovely"/ but it is set to "  Import \"spam\"\n\n  <Module \"spam\">\n    spam \"\"wonderful\" \"lovely\"\"\n  </Module>\n"
       Diff:
       @@ -1,2 +1,6 @@
       -(?-mix:spam "wonderful" "lovely")
       +  Import "spam"
       +
       +  <Module "spam">
       +    spam ""wonderful" "lovely""
       +  </Module>

     # ./spec/defines/collectd_plugin_python_module_spec.rb:77:in `block (3 levels) in <top (required)>'
```

So I modified the test case to look for a value which is a string of quoted words and ensure the output is still valid.

I'm not convinced this will solve the nested data structure stuff - got an example?

jyaworski: @igalic had suggested shellwords as a way to quote things without knowing if they had quotes around them. For some reason it's not working under jruby in particular.

jyaworski:
@bilsch please rebase against master to get the new travis config.

templates/plugin/python/module.conf.erb
bilsch: tests passed. Should I revert the modules.conf.erb back to current master and push to trigger the builds again? Curious if the jruby buld slaves catch the failures ( would leave my tests in place )

jyaworski: It's worth a shot.

bilsch:
curious that the 2.3 ruby worked fine but the others all failed.

Thoughts on using shellwords - seems ok but only on more recent rubies.

jyaworski:
This wouldn't be the first time that the ruby version has produced absolutely bizarre results. If you can replicate the functionality without using it, I'm fine with it.

bilsch:
added WIP as I'm still working on something else which needs this change. I'll get the changes discussed with @jyaworski here in a little bit - probably tomorrow

jyaworski:
@bilsch did you see https://github.com/puppetlabs/puppetlabs-stdlib/pull/580?

It seems like 'require shellwords' works.

bilsch:
nifty good find @jyaworski . I'm on vacation / moving this coming week. I'll try to tinker with this a little bit this week but may not get to it for a bit

jyaworski:
@bilsch I fixed the python issue in itself. Can you open a new PR or change this one and rebase with the other changes?

igalic:
jruby and shell don't seem to be friends :(

PR #480, opened 2016-04-26
Add options VerifyPeer, VerifyHost and CaCert to curl_json plugin

matejzero:
It seems like those options are available in curl plugin, but not curl_json.


bastelfreak:
Hi, could you please add spec tests to the changes? I think there are none for curl_json :cry:

matejzero:
I added some tests to the ones that were already in place. I have to say I'm no good at writing tests, so I just copy/paste and change so it makes sense to me:)

matejzero:
I hope this is enough.

spec/defines/collectd_plugin_curl_json_spec.rb
```diff
@@ -13,6 +13,9 @@
     {
       url: 'http://localhost:55672/api/overview',
       instance: 'rabbitmq_overview',
+      verifypeer: 'false',
+      verifyhost: 'false',
```
bastelfreak: I am not sure if 'false' here equals the default undef in the manifest.

CC: @jyaworski

bastelfreak:
I'm not sure about the 'false' parts, the rest looks good, thanks for that. I would like to keep this PR on hold until we do a modulesync (hopefully in the next few hours or tomorrow). Then the testsuite for ruby<2.2 will work again.

spec/defines/collectd_plugin_curl_json_spec.rb
matejzero: False doesn't equal undef in manifest. If you don't set those variables, puppet won't set this parameter in collectd config and collectd will use the default, which is true for those two parameters. Should I change that to true?

matejzero:
No problem, take your time.

jyaworski:
@matejzero please rebase against master and squash.

spec/defines/collectd_plugin_curl_json_spec.rb
jyaworski: @bastelfreak 'false' might not, but false does I believe.

matejzero:
I'm not sure if I did that right. Still a GIT noob, I will have to do some trial/error on private repo:)

jyaworski:
@matejzero here are a couple of examples:

https://help.github.com/articles/about-git-rebase/
https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history

Essentially the goal is to rebase and squash. Rebasing is making sure your PR is up-to-date with the master code, and squashing is removing 'dumb' commits like 'fix 1', 'fix2' and so on. In the end you have a few very good commits.

matejzero:
I think I did the squash correctly, but I'm not sure about the rebase with upstream.

PR #482, opened 2016-04-27
initial spec tests for write_riemann plugin class

archii:


archii:
Need to work on this some more after rubocop errors..

PR #484, opened 2016-04-27
initial spec tests for write_riemann plugin class

archii:


bastelfreak:
Thanks for the contribution!

PR #487, opened 2016-04-29
Check out what using rubocop-rspec does

jyaworski:
@bastelfreak @rnelson0

### DIFF
--- a/spec/classes/collectd_init_spec.rb
+++ b/spec/classes/collectd_init_spec.rb
@@ -32,7 +32,7 @@ describe 'collectd' do
         { package_install_options: 'not_an_array' }
       end
 
-      it 'should fail' do
+      it 'fails' do
         expect do
           should contain_class('collectd')
         end.to raise_error(Puppet::Error, /is not an Array/)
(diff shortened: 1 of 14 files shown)


jyaworski:
OK. @bastelfreak @rnelson0 @nibalizer

This is what autocorrect gave me. Do we want to disable the name cop?

rnelson0:
It looks okay to me, but I definitely have opinions on word tenses in commit messages, for instance.

jyaworski:
@rnelson0 it specifically says something like 'do not use should to describe'....

Any opinions on disabling cops?

spec/classes/collectd_init_spec.rb
```diff
@@ -32,7 +32,7 @@
         { package_install_options: 'not_an_array' }
       end
 
-      it 'should fail' do
+      it 'fails' do
```
igalic: i like the optimism in these changes ;)

jyaworski: I do too. I think we should enable this.

PR #489, opened 2016-04-29
add write_kafka support

bnied:
This PR adds support for the [write_kafka](https://collectd.org/wiki/index.php/Plugin:Write_Kafka) plugin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,7 @@ documentation for each plugin for configurable attributes.
 * `vmem` (see [collectd::plugin::vmem](#class-collectdpluginvmem) below)
 * `write_graphite` (see [collectd::plugin::write_graphite](#class-collectdpluginwrite_graphite) below)
 * `write_http` (see [collectd::plugin::write_http](#class-collectdpluginwrite_http) below)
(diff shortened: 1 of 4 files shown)


jyaworski:
Does this require an external package?

Also, maybe note it's in 5.5+.

bnied:
Yes, this requires `librdkafka1` to be present.

I can note the version requirement; do you want this in the README?

jyaworski:
@bnied yeah, that looks good. I meant, for example, the disk plugin which on RHEL as 'collectd-disk.' External requirements like that can just be noted. It's the same way for Redis.

bnied:
Gotcha. On Ubuntu, the plugin should be included in the collectd or collectd-core packages.

On Fedora/RHEL, the package will be `collectd-write_kafka` (going off of [their official specfile](https://github.com/collectd/collectd/blob/master/contrib/redhat/collectd.spec#L764-L772)), though it appears that neither RHEL/CentOS or Fedora package the library or the write_kafka plugin.

bnied:
@jyaworski added the notes to the README. Is there anything else you need for this PR?

bnied:
Thanks!

PR #494, opened 2016-05-04
add spec tests, update README, fix template for write_tsdb

mothbe:


mothbe:
@jyaworski now it looks good

PR #495, opened 2016-05-06
Add ExcludeRegex option to tail module

matejzero:
Tail plugin has an option to also set ExcludeRegex.

I tried adding a test as well, but I'm having some troubles. Will try some more and commit when I'll have something useful.
I also can't find a good case for documentation.

### DIFF
--- a/templates/tail-file.conf.erb
+++ b/templates/tail-file.conf.erb
@@ -6,6 +6,7 @@
 <% @matches.each do |match| -%>
   <Match>
    Regex "<%= match['regex'] %>"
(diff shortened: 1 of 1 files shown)


jyaworski:
Hello @matejzero.

Do you know in what version this was added? If so, please wrap the line in a version conditional.

matejzero:
It was added in 4.10.0 and 5.1.0.
I checked supported distros and they all have those versions and newer. Is version check still necessary in that case?

igalic:
i'm okay with this, if you are, @jyaworski

matejzero:
Great! Thanks for merge.

PR #500, opened 2016-05-09
Fix rabbitmq module

bilsch:
This is picking up where https://github.com/voxpupuli/puppet-collectd/pull/478 left off. I have reduced the changes in this to just what is needed to fix my busted stuff for the rabbitmq python module check.

I'm having problems getting the rake tasks to run due to a gem issue. Curious if this is occurring upstream as well.

```
An error occurred while installing listen (3.1.3), and Bundler cannot continue.
Make sure that `gem install listen -v '3.1.3'` succeeds before bundling.
```

The tests were passing previously in #478 so I believe this will pass

### DIFF
--- a/spec/classes/collectd_plugin_rabbitmq_spec.rb
+++ b/spec/classes/collectd_plugin_rabbitmq_spec.rb
@@ -11,6 +11,18 @@ describe 'collectd::plugin::rabbitmq', type: :class do
   context 'package ensure' do
     context ':ensure => present' do
       let(:node) { 'testhost.example.com' }
+      let :params do
+        {
+          config: {
+            'Username' => 'guest',
+            'Password' => 'guest',
+            'Port'     => '15672',
+            'Scheme'   => 'http',
+            'Host'     => 'testhost.example.com',
+            'Realm'    => '"RabbitMQ Management"'
+          }
+        }
+      end
 
       it 'Load collectd_rabbitmq in python-config' do
         should contain_concat_fragment('collectd_plugin_python_conf_collectd_rabbitmq.collectd_plugin').with_content(/Module "collectd_rabbitmq.collectd_plugin"/)
@@ -38,6 +50,10 @@ describe 'collectd::plugin::rabbitmq', type: :class do
 
       it 'Host should be set to $::fqdn python-config' do
         should contain_concat_fragment('collectd_plugin_python_conf_collectd_rabbitmq.collectd_plugin').with_content(/Host "testhost.example.com"/)
+      end
+
+      it 'Realm set to "RabbitMQ Management"' do
+        should contain_concat_fragment('collectd_plugin_python_conf_collectd_rabbitmq.collectd_plugin').with_content(/Realm "RabbitMQ Management"/)
       end
     end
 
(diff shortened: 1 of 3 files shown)


jyaworski:
It looks like a formatting problem, here.

bilsch:
😦 not sure  #490 actually worked. All test cases failed with quoted words instead of a single quoted string.

```
Failures:
  1) collectd::plugin::rabbitmq package ensure :ensure => present Realm set to "RabbitMQ Management"
     Failure/Error: should contain_concat_fragment('collectd_plugin_python_conf_collectd_rabbitmq.collectd_plugin').with_content(/Realm "RabbitMQ Management"/)
       expected that the catalogue would contain Concat_fragment[collectd_plugin_python_conf_collectd_rabbitmq.collectd_plugin] with content set to /Realm "RabbitMQ Management"/ but it is set to "  Import \"collectd_rabbitmq.collectd_plugin\"\n\n  <Module \"collectd_rabbitmq.collectd_plugin\">\n    Host \"testhost.example.com\"\n    Password \"guest\"\n    Port \"15672\"\n    Realm \"RabbitMQ\" \"Management\"\n    Scheme \"http\"\n    Username \"guest\"\n  </Module>\n"
       Diff:
       @@ -1,2 +1,11 @@
       -(?-mix:Realm "RabbitMQ Management")
       +  Import "collectd_rabbitmq.collectd_plugin"
       +
       +  <Module "collectd_rabbitmq.collectd_plugin">
       +    Host "testhost.example.com"
       +    Password "guest"
       +    Port "15672"
       +    Realm "RabbitMQ" "Management"
       +    Scheme "http"
       +    Username "guest"
       +  </Module>
     # ./spec/classes/collectd_plugin_rabbitmq_spec.rb:56:in `(root)'
```

spec/classes/collectd_plugin_rabbitmq_spec.rb
```diff
@@ -11,6 +11,18 @@
   context 'package ensure' do
     context ':ensure => present' do
       let(:node) { 'testhost.example.com' }
+      let :params do
+        {
+          config: {
+            'Username' => 'guest',
```
jyaworski: Can you align these =>?

jyaworski:
Oh, I see @bilsch. Yeah, I ran into that. I had to double quote it....

```yaml
collectd::plugin::rabbitmq::config:
  'Username': 'nagios'
  'Password': 'blah'
  'Scheme': 'http'
  'Port': '15672'
  'Host': "%{::fqdn}"
  'Realm': '"RabbitMQ Management"'
```

I agree it's not a good solution, but it did 'work.'

bilsch:
@jyaworski ah ok I can do that. Will push up a change in just a sec

bilsch:
yea, this looks better

```
+++ /tmp/puppet-file20160509-26748-141keft	2016-05-09 16:25:08.580363134 -0400
@@ -10,7 +10,7 @@
     Host "snip"
     Password "snip"
     Port "15671"
-    Realm "RabbitMQ" "Management"
+    Realm "RabbitMQ Management"
     Scheme "https"
     Username "snip"
   </Module>
```

I think WIP can finally be taken off of this one! Thanks for your help on this @jyaworski !

bilsch:
one last little tweak to the readme for the quoting. Assuming those pass ( minus 2.3.0 ruby ) imo this is finally done

PR #501, opened 2016-05-10
Update from voxpupuli modulesync_config

jyaworski:


jyaworski:
Calling ruby experts @rnelson0 and @igalic.

jyaworski:
Looks like here:

https://stackoverflow.com/questions/32962161/rspec-given-error-then-is-not-available-from-within-an-example-e-g-an-it-b

jyaworski:
Do we need tmpfilename here?

rnelson0:
It looks like the remove of `spec/spec_helper.rb` line 4 removed that method. I always use `:concat_basedir => '/dne',` (dne == 'does not exist') as it's building the catalog, not actually implementing it, so I think that's probably the best way forward.

jyaworski:
@rnelson0 done.

rnelson0:
Awesome, all tests pass now!

PR #502, opened 2016-05-15
Adds the ability to set tags and attributes for the `write_riemann` plugin

torrancew:
This has been tested in a real environment. Unit tests are also provided.

I'm a bit on the fence about calling the parameter `tags`, due to possible confusion with the Puppet metaparameter `tag`, but tried to follow existing conventions from this and other plugin manifests.


bastelfreak:
thanks for the contribution!

PR #503, opened 2016-05-18
SNMP Plugin - Instance should be optional

Lavaburn:
The SNMP plugin allows Instance to be undefined if Table is set to true.

This is documented on https://collectd.org/documentation/manpages/collectd-snmp.5.shtml
"If Table is set to true and Instance is omitted, then "SUBID" will be used as the instance."

### DIFF
--- a/manifests/plugin/curl_json.pp
+++ b/manifests/plugin/curl_json.pp
@@ -27,9 +27,7 @@ define collectd::plugin::curl_json (
     }
 
     if $::osfamily == 'Redhat' {
-      package { 'collectd-curl_json':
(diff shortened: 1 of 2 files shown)


jyaworski:
Linting doesn't like you. Please fix that. Ignore the rubocop errors, unless you feel like fixing them. It seems like rubocop -a and puppet-lint -f will do it.

jyaworski:
@Lavaburn can you rebase? We made a few spec changes.

Lavaburn:
Thanks. Seems to pass now.

PR #507, opened 2016-05-24
Only set ExcludeRegex in config file if variable is defined

matejzero:
ExcludeRegex was set in config file even if it wasn't defined. This fix only sets ExcludeRegex when you define it in manifest.


matejzero:
You might want to pull that merge back. The solution doesn't seem to work on my test server and I can't see the problem. Will do more testing tomorrow and report back.

bastelfreak:
reverted in https://github.com/voxpupuli/puppet-collectd/pull/508

PR #509, opened 2016-05-25
Only set ExcludeRegex when needed (fix for #507)

matejzero:
Previous if statement was wrong, now it's working as it should...

### DIFF
--- a/templates/tail-file.conf.erb
+++ b/templates/tail-file.conf.erb
@@ -6,7 +6,9 @@
 <% @matches.each do |match| -%>
   <Match>
    Regex "<%= match['regex'] %>"
+<% if match['excluderegex'] -%>
    ExcludeRegex "<%= match['excluderegex'] %>"
+<% end -%>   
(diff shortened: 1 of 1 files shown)


templates/tail-file.conf.erb
```diff
@@ -6,7 +6,9 @@
 <% @matches.each do |match| -%>
   <Match>
    Regex "<%= match['regex'] %>"
+<% if match['excluderegex'] -%>
    ExcludeRegex "<%= match['excluderegex'] %>"
+<% end -%>
```
alexjfisher: nitpick: trailing whitespace

alexjfisher: which results in extra 'blank' lines in the outputted file...
https://travis-ci.org/voxpupuli/puppet-collectd/jobs/132898797#L1607

matejzero:
Sorry about the trailing whitespaces. Usually sublime takes care of that but somehow it didn't. I see @alexjfisher already fixed this.

PR #513, opened 2016-05-27
Check $manage_package if not undef

jkroepke:
If $manage_package is false, the else block will be used instead disable managing the package.

### DIFF
--- a/manifests/plugin/disk.pp
+++ b/manifests/plugin/disk.pp
@@ -15,7 +15,7 @@ class collectd::plugin::disk (
   validate_bool($ignoreselected)
 
   if $::osfamily == 'RedHat' {
-    if $manage_package {
+    if $manage_package != undef {
       $_manage_package = $manage_package
     } else {
       if versioncmp($::collectd::collectd_version_real, '5.5') >= 0 {
--- a/spec/classes/collectd_plugin_disk_spec.rb
+++ b/spec/classes/collectd_plugin_disk_spec.rb
@@ -48,6 +48,25 @@ describe 'collectd::plugin::disk', type: :class do
     end
   end
 
+  context ':manage_package => false on osfamily => RedHat' do
+    let :facts do
+      {
+        osfamily: 'RedHat',
+        collectd_version: '5.5',
+      }
+    end
+
+    let :params do
+      {
+        manage_package: false,
+      }
+    end
+    it 'Will not manage collectd-disk' do
+      should_not contain_package('collectd-disk').with(ensure: 'present',
+                                                       name: 'collectd-disk',)
+    end
+  end
+
   context ':manage_package => undef on osfamily => RedHat with collectd 5.5 and up' do
     let :facts do
       {


manifests/plugin/disk.pp
```diff
@@ -15,7 +15,7 @@
   validate_bool($ignoreselected)
 
   if $::osfamily == 'RedHat' {
-    if $manage_package {
+    if $manage_package != undef {
```
alexjfisher: This looks correct.  Perhaps throw in a `validate_bool($manage_package)`?

jkroepke: Like 

```
if $manage_package != undef {
  validate_bool($manage_package)
}
```

?

alexjfisher: may as well.  It'd guard against people using the string 'false' (which is true!).
If you're feeling really generous, a spec test would be great, but don't worry if you haven't got the time.  Thanks for spotting this bug!

jkroepke: Write 'false' in hiera result into false, not true. I don't have great knowledge on spec test but I will try it.

spec/classes/collectd_plugin_disk_spec.rb
```diff
@@ -48,6 +48,25 @@
...
+      }
+    end
+
+    let :params do
+      {
+        manage_package: false,
+      }
+    end
+    it 'Will not manage collectd-disk' do
+      should_not contain_package('collectd-disk').with(ensure: 'present',
```
alexjfisher: You can probably leave off the `.with` bit here as we don't care what the parameters are; just that the catalog doesn't contain any package resource for 'collectd-disk'

jkroepke: Thanks for the info. I copied it from here: https://github.com/voxpupuli/puppet-collectd/blob/master/spec/classes/collectd_plugin_disk_spec.rb#L73 :)

manifests/plugin/disk.pp
alexjfisher: Curious... Maybe it's dependent on the version of puppet?

```
alex@alex-VirtualBox:~$ cat test.pp
if 'false' {
  notice("'false' is true")
}
if false {
  notice("false is true??")
} else {
  notice("but false is false!")
}
alex@alex-VirtualBox:~$ puppet apply test.pp
Notice: Scope(Class[main]): 'false' is true
Notice: Scope(Class[main]): but false is false!
Notice: Compiled catalog for alex-virtualbox.ordsvy.gov.uk in environment production in 0.07 seconds
Notice: Applied catalog in 0.01 seconds
alex@alex-VirtualBox:~$ puppet --version
4.5.0
```

Anyway, congrats on adding a spec test!  I suppose you could add a context with `:manage_package => 'false'`  Hopefully watch it fail (on some versions of puppet only perhaps).  Then add a validate_bool and change the expectation to something similar to the test in `context ':disks is not an array'`

But only if you're bored and want to do some playing around with rspec-puppet! ;) If you're short of time, I'm happy with merging without the extra tests.

spec/classes/collectd_plugin_disk_spec.rb
alexjfisher: Feel free to tidy that up too! ;)
https://xkcd.com/386/

jkroepke: Great comic! Love this side.

manifests/plugin/disk.pp
jyaworski: This can just be 'if $manage_package'

jyaworski:
I think we'll solve this with a manage_repo parameter.

markasammut:
Can this bug fix be merged in a new release please? It is an absolute necessity in RH7 hosts installing collectd from epel

eoly:
+1 for merging and publishing a new release.

This bug is preventing me from upgrading from the pdxcat version of the collectd module to this one. We are installing collectd from the epel repos on Cent 7. I imagine this is a fairly common use case.

jyaworski:
I merged it; if someone cares enough we can revisit this. It solves a problem people have now, though.

eoly:
:+1: thanks!

jkroepke:
Thanks! Now we can upgrade.

PR #516, opened 2016-05-31
Add spec test for 'has_wordexp' and validate_bool

alexjfisher:
Also add validation for most of the other parameters.


alexjfisher:
Spec test for https://github.com/voxpupuli/puppet-collectd/pull/515

PR #517, opened 2016-06-01
Apache plugin - Remove quotes from VerifyPeer and VerifyHost

dansajner:
Currently the plugin fails to load when passing `verifypeer` and `verifyhost` args to an instance definition:
`cf_util_get_boolean: The VerifyPeer option requires exactly one boolean argument.`


jyaworski:
@dansajner looks good. Can you squash at least the last two commits together?

dansajner:
@jyaworski squashed to 1 commit.

alexjfisher:
Is it worth adding
```
if $verifypeer != undef {
  validate_bool($verifypeer)
}
if $verifyhost != undef {
  validate_bool($verifyhost)
}
```
to https://github.com/voxpupuli/puppet-collectd/blob/master/manifests/plugin/apache/instance.pp ?

This would make it impossible to accidentally pass incorrect values through to the template.

PR #518, opened 2016-06-07
target replace filter fix to allow identical keys with different values

gregsaram:
This change allows having several occurrences of PluginInstance within the same target filter definition block.

```
collectd::plugin::filter::target{ "overwrite OIDs":
      chain   => $chainname,
      rule    => $rulename,
      plugin  => 'replace',
      options => {
          'PluginInstance' => ['"string1." "replace_string_1_"','"string2-" "replace_string_2"','"\"" "."']
      }
```

Resulting in this output in config file
```
<Target "replace">
   PluginInstance "string1." "replace_string_1_"
   PluginInstance "string2-" "replace_string_2_"
   PluginInstance "\"" "."
</Target>
```


jyaworski:
Can you add some context as to why this change is useful? What could be done if we merged this?

PR #532, opened 2016-07-13
add ability to add files via hiera

sigsegv0x0b:
ex:

    collectd::plugin::tail::files:
      'somelog':
        filename: '/var/log/logfile'
        instance: 'lookingforsomething'
        matches:
          - regex:    'connected'
            dstype:   'CounterInc'
            type:     'counter'
            instance: 'connection-open'
          - regex:    'disconnect'
            dstype:   'CounterInc'
            type:     'counter'
            instance: 'connection-close'

### DIFF
--- a/manifests/plugin/tail.pp
+++ b/manifests/plugin/tail.pp
@@ -2,8 +2,16 @@
 # https://collectd.org/wiki/index.php/Plugin:Tail
 class collectd::plugin::tail (
   $interval = undef,
+  $files = undef,
 ) {
+
+  if $files { validate_hash($files) }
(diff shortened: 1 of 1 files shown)


bastelfreak:
Hi @sigsegv0x0b, thans for the PR!
Two things:
* Could you add your hiera example to the README.md?
* Could you add tests for your changes?

sigsegv0x0b:
This doesn't add anything new, it just fixes what was most likely missing create resources from tail.pp. So current documentation and tests should really cover it same way as they do now.

PR #546, opened 2016-08-30
add libyajl2 dep for ceph

wasapi:
Not so necessary, add libyajl lib for ceph plugin, like curl_json plugin.


wasapi:
Not so good...

Issue #552, opened 2016-09-23
CentOS 6/7 no package collectd


Yuav:
### Affected Puppet, OS and module versions/distributions

- Puppet: 4.7.0
- Distribution: CentOS 6 and CentOS 7
- Module version: 5.1.0

### How to reproduce (e.g Puppet code you use)

class { 'collectd': }

### What are you seeing

    [root@localhost /]# yum install collectd
    Loaded plugins: fastestmirror, ovl
    Setting up Install Process
    Loading mirror speeds from cached hostfile
     * base: mirror.colocenter.nl
     * extras: mirrors.noction.com
     * updates: mirror.proserve.nl
    No package collectd available.
    Error: Nothing to do

### Output log

    Error: Execution of '/usr/bin/yum -d 0 -e 0 -y install collectd' returned 1: Error: Nothing to do
    Error: /Stage[main]/Collectd::Install/Package[collectd]/ensure: change from purged to present failed:         
    Execution of '/usr/bin/yum -d 0 -e 0 -y install collectd' returned 1: Error: Nothing to do

PR #556, opened 2016-09-23
Fix for issue #552 including acceptance tests

Yuav:
Adding Docker nodesets for
Debian 7/8
Ubuntu 12/14/16
Ommiting CentOS 6/7 due to bug

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -5,5 +5,7 @@ fixtures:
     concat:
       repo: 'git://github.com/puppetlabs/puppetlabs-concat'
       ref: '2.1.0'
+    yum:
+      repo: 'https://github.com/voxpupuli/puppet-yum.git'
   symlinks:
     'collectd': "#{source_dir}"
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -11,6 +11,15 @@ class collectd::install (
   }
 
   if $manage_package {
+    if !defined(Yum::Install['epel-release']) {
+      if $::osfamily == 'RedHat' {
+	      yum::install { 'epel-release':
+	        ensure => 'present',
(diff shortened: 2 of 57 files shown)


Yuav:
Seems Travis needs to retry on Debian

Yuav:
Retrying

Yuav:
Fixes #552 with acceptance tests to prove it

daenney:
This is way way way too many commits. Please do some work to rewrite history in much more processable and contained sets of commits.

Also, the changes to `.travis.yml` might get module synced away.

Yuav:
I did a rebase from master. Not sure why they all ended up in the log

daenney:
I'm not sure the Travis+Docker changes is something we want to do right now. This is going to dramatically slow down tests, it's adding  9 tests to the matrix and Beaker tests are pretty slow in general.

Yuav:
The module was unusable on CentOS 6 and 7 for instance. The added tests would ensure that doesn't happen again

Yuav:
Not sure I agree with "dramatically" though? Spec test on Ruby 2.3.1 is already 4 minutes, which is longer than most of the Beaker tests. In my opinion the Beaker tests have more value than multiple ruby versions

daenney:
It's at worst 9x~3m that get added to the test time when it's busy and things can't get scheduled in parallel, That's almost 20 minutes extra wait time, in practice it's probably around 10m.

Yuav:
In my opinion it's still worth it :-)

bastelfreak:
IMO we should wait with the acceptance test until https://github.com/voxpupuli/modulesync_config/pull/230 is resolved. Maybe it is a good idea to add the tests from this PR in a new one?

daenney:
The adding and revert should be squashed to just eliminate them from history.

PR #557, opened 2016-09-29
Added missing quote

Yuav:


bastelfreak:
Thanks for the fix @Yuav!

PR #561, opened 2016-10-13
Added provider_proxy for pip

nurriz:

### DIFF
--- a/manifests/plugin/rabbitmq.pp
+++ b/manifests/plugin/rabbitmq.pp
@@ -69,7 +69,7 @@ class collectd::plugin::rabbitmq (
     $install_options = [{'--proxy' => $provider_proxy}]
     else {
     $install_options = undef
-    }    
+    }
     package { $package_name:
       ensure          => $ensure,
       provider        => $package_provider,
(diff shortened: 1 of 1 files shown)


manifests/plugin/rabbitmq.pp
```diff
@@ -64,10 +65,17 @@
 
   $_manage_package = pick($manage_package, $::collectd::manage_package)
 
-  if $_manage_package {
+  if ($_manage_package) and ($provider_proxy) {
```
bastelfreak: can you here only set $install_options to `[{'--proxy' => $provider_proxy}]` and in the else part to `undef`?

This allows us to only define the package resource once.

nurriz: Is this what you were thinking?

bastelfreak:
yep, thats what I meant. Can you please add tests as well?

manifests/plugin/rabbitmq.pp
```diff
@@ -64,12 +65,17 @@
 
   $_manage_package = pick($manage_package, $::collectd::manage_package)
 
-  if $_manage_package {
-    package { $package_name:
-      ensure   => $ensure,
-      provider => $package_provider,
-    }
+  if ($_manage_package) and ($provider_proxy) {
```
jyaworski: Why are these in parentheses?

bastelfreak: they aren't needed, but I'm fine with them.

PR #568, opened 2016-10-19
Add apt repo

Yuav:
Get latest version of collectd including plugins

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -1,5 +1,7 @@
 fixtures:
   repositories:
+    apt:
+      repo: 'git://github.com/puppetlabs/puppetlabs-apt'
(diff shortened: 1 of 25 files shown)


bastelfreak:
Hi, thanks for providing this PR. Two things:
* Could you add add a parameter like $manage_repo so users can decide if they want to have the ppa or not?
* Instead of mocking more facts, could you migrate the tests to rspec-puppet-facts? I did that already for a few other tests here: https://github.com/voxpupuli/puppet-collectd/pull/567

Yuav:
1. Fixed
2. Was fairly few plugins missing this particular fact. I agree it's needed, but slightly out of scope for this patch?

PR #571, opened 2016-10-21
Add support for ReportNumCpu in CPU plugin introduced in collectd 5.6

matejzero:
As title says, this PR adds support for ReportNumCpu. From collectd site: When set to true, reports the number of available CPUs.


jyaworski:
Please rebase.

PR #579, opened 2016-11-08
Ensure ppa is applied before package is installed

Yuav:
Missing ordering logic for ppa on Ubuntu

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -35,6 +35,11 @@ class collectd::install (
             'server' => 'keyserver.ubuntu.com',
           },
         }
+        if defined(Package[$package_name]) {
+          Package[$package_name] {
+            require => Apt::Source['ppa_collectd']
+          }
+        }
       }
       default                   : {
       }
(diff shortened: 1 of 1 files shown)


dhoppe:
@Yuav I think we need to refactor this manifest. First we should configure the repository/ppa, based on `manage_repo` and then we should install the package based on `manage_package`, where we store the requirement as variable and finally use it.

manifests/install.pp
```diff
@@ -17,26 +17,20 @@
...
-      '12.04', '14.04', '16.04' : {
-        apt::source { 'ppa_collectd':
-          location => 'http://ppa.launchpad.net/collectd/collectd-5.5/ubuntu',
-          repos    => 'main',
-          key      => {
-            'id'     => '7543C08D555DC473B9270ACDAF7ECBB3476ACEB3',
-            'server' => 'keyserver.ubuntu.com',
-          },
-        }
+    } elsif ($::operatingsystemmajrelease == '12.04' or $::operatingsystemmajrelease == '14.04' or $::operatingsystemmajrelease == '16.04') {
```
dhoppe: This looks better than my approach, but I would use `$::operatingsystemmajrelease =~ /('12.04'|'14.04'|'16.04')/` instead of these long line.

jyaworski: `(12|14|16).04`?

dhoppe: Even better ;)

Yuav: Agree!

PR #582, opened 2016-11-15
collectd::plugin::amqp is now supporting the routingkey option

alexxxxx:
Trivial pull request adding support of the amqp option "routingkey"


bastelfreak:
Hi @alexxxxx, thanks for the PR. Are you able to provide tests as well?

oranenj:
I added some base for the tests in #661.

Adding more for these will be easy once that's merged.

Issue #550, opened 2016-09-23
Version definition ineffective if plugins are installed


markasammut:
### Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 3.8.3
- Ruby: 1.8.7
- Distribution: RHEL 7
- Module version: 5.1.0

### What are you seeing

When a collectd plugin is specified to be installed (like lvm), it doesn't use the same package version as the collectd package, but uses the latest version, with the result that the collectd package is upgraded due to it being a dependency.

### What behaviour did you expect instead

I'd expect the plugins to use the same version as the main collectd.

markasammut:
Solved by https://github.com/voxpupuli/puppet-collectd/pull/551 for disk, lvm and python plugins. Changes need to be replicated for other plugins.

PR #589, opened 2016-11-30
ensure same version as collectd

markasammut:


bastelfreak:
@markasammut thanks for the PR! Can you take a look at the failing tests?

You can always join us in #voxpupuli on freenode (IRC) if you've any questions about testing.

juniorsysadmin:
@markasammut It would be great if there was a full Pull Request description describing what you are attempting to do and what it fixes

markasammut:
This is the same issue as #550  and #551

markasammut:
As things stand, if you specify a version which is not the latest for collectd, a conflict will be created if a plugin like curl is added. This is because curl will try to be installed with the latest version (and nowhere in the manifest does it instruct it to use the same version as collectd). Subsequent puppet runs will fail as there's an rpm dependency between collectd and the plugins, and collectd will have been updated to the latest package and unable to be downgraded due to a conflict with the plugin's latest version.

This pull request attempts to instruct the plugin to use the same version as collectd if ensure=present.

markasammut:
Can this be re-tested now?

bastelfreak:
Hi @markasammut, your branch has a merge conflict. Can you rebase against our latest master (rebase, not merge)? Feel free to ping us on the freenode IRC channel #voxpupuli if you need any help with git or testing.

bastelfreak:
Thanks for bringing up this PR. I'm going to close this because of inactivity. Please reopen it if you're still interested in it.

PR #594, opened 2016-12-08
Add rabbitmq types to collectd types database on install

Yuav:
Rabbitmq plugin use custom types that needs to be added to CollectD types database for collection to work

### DIFF
--- a/manifests/plugin/rabbitmq.pp
+++ b/manifests/plugin/rabbitmq.pp
@@ -97,6 +97,15 @@ class collectd::plugin::rabbitmq (
     install_options => $install_options,
   }
 
+  $rt = '/usr/local/share/collectd-rabbitmq/types.db.custom' # RabbitMQ addon types file
+  $ct = '/usr/share/collectd/types.db'                       # CollectD types file
+
+  exec { 'update_typesdb':
+    command => "/bin/cat ${rt} >> ${ct}",
+    unless  => "[ `cat ${ct} | sort -u | wc -l` -eq `cat ${rt} ${ct} | sort -u | wc -l` ] || exit 1",
+    path    => ['/usr/bin', '/bin'],
+  }
+
   collectd::plugin::python::module { 'collectd_rabbitmq.collectd_plugin':
     ensure => $ensure,
     config => $config,
--- a/manifests/plugin/rabbitmq.pp
+++ b/manifests/plugin/rabbitmq.pp
@@ -101,9 +101,10 @@ class collectd::plugin::rabbitmq (
   $ct = '/usr/share/collectd/types.db'                       # CollectD types file
 
   exec { 'update_typesdb':
-    command => "/bin/cat ${rt} >> ${ct}",
-    unless  => "true && case $(cat ${ct}) in *$(cat ${rt})*) exit 0;; *) exit 1;; esac",
-    path    => ['/usr/bin', '/bin'],
+    command  => "/bin/cat ${rt} >> ${ct}",
+    unless   => "case $(cat ${ct}) in *$(cat ${rt})*) exit 0;; *) exit 1;; esac",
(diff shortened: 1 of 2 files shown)


manifests/plugin/rabbitmq.pp
```diff
@@ -97,6 +97,15 @@
     install_options => $install_options,
   }
 
+  $rt = '/usr/local/share/collectd-rabbitmq/types.db.custom' # RabbitMQ addon types file
+  $ct = '/usr/share/collectd/types.db'                       # CollectD types file
+
+  exec { 'update_typesdb':
+    command => "/bin/cat ${rt} >> ${ct}",
+    unless  => "[ `cat ${ct} | sort -u | wc -l` -eq `cat ${rt} ${ct} | sort -u | wc -l` ] || exit 1",
```
bastelfreak: I need to think a bit how we can make this a bit more beautiful

bastelfreak:
Thanks for the PR @Yuav! Are you able to also add unit tests for the change?

manifests/plugin/rabbitmq.pp
Yuav: I stole this approach from: https://stackoverflow.com/questions/16349543/how-to-determine-if-the-content-of-one-file-is-included-in-the-content-of-anothe

Yuav:
I'm not sure what that unit test should look like?

Testing like this looks useless to me:

    is_expected.to contain_exec('update_typesdb')

manifests/plugin/rabbitmq.pp
```diff
@@ -102,7 +102,7 @@
 
   exec { 'update_typesdb':
     command => "/bin/cat ${rt} >> ${ct}",
-    unless  => "[ `cat ${ct} | sort -u | wc -l` -eq `cat ${rt} ${ct} | sort -u | wc -l` ] || exit 1",
+    unless  => "true && case $(cat ${ct}) in *$(cat ${rt})*) exit 0;; *) exit 1;; esac",
```
alexjfisher: @Yuav Hack looks hacky.  Did you try `provider => shell`? https://docs.puppet.com/puppet/latest/types/exec.html#exec-provider-shell

Yuav: Provider => shell is better. Fixed.

manifests/plugin/rabbitmq.pp
```diff
@@ -97,6 +97,16 @@
...
 
+  $rt = '/usr/local/share/collectd-rabbitmq/types.db.custom' # RabbitMQ addon types file
+  $ct = '/usr/share/collectd/types.db'                       # CollectD types file
+
+  exec { 'update_typesdb':
+    command  => "/bin/cat ${rt} >> ${ct}",
+    unless   => "case $(cat ${ct}) in *$(cat ${rt})*) exit 0;; *) exit 1;; esac",
+    path     => ['/usr/bin', '/bin'],
+    provider => 'shell',
+  }
```
igalic: this whole thing should be a `file_line` resource.

igalic:
that is, if my understanding here is correct, then file_line is the way to go, if not, then perhaps we should make `$rt = '/usr/local/share/collectd-rabbitmq/types.db.custom'` a template / `concat`?

Yuav:
The rabbitmq plugin doesn't work at all now though. As far as I can tell it never has.

I wanted to use the upstream files to avoid having to update template on every plugin and/or collectd release. As far as I know, there's no collisions between the files as the counters are specific to the plugin.

As new plugins and features essentially gets added to types.db on every release, both file_line and concat would enforce hardcoding types.db contents in template or .pp.

The better approach is to use the TypesDB option in collectd.conf to include both files. That rewrite is not really a good candidate for this PR.

igalic:
aye.
thank you very much for this research @Yuav!

PR #597, opened 2016-12-12
Include custom rabbitmq types

Yuav:
Required for the RabbitMQ plugin to work


traylenator:
I think it's fine in principle, I just restarted the job that failed...

petems:
The typesdb change is changed/breaking behaviour, so this'll have to have a major version bump.

PR #599, opened 2016-12-14
Update fixtures

petems:
* Make sure we're specifying versions from metadata

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -2,12 +2,15 @@ fixtures:
   repositories:
     apt:
       repo: 'git://github.com/puppetlabs/puppetlabs-apt'
+      tag: '2.2.0'
     stdlib:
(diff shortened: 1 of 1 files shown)


bastelfreak:
mhm, I'm not sure if we want that. In general we try to test against master to detect unreleased issues in our dependcies. This revealed several stdlib issues that I'm aware of and probably more. On the negative site we never know if our dependencies in metadata.json are working with the module or if they are outdated.

I would still prefer option 1 for all the unit tests. We can do real testing if correct dependencies in our acceptance tests.

juniorsysadmin:
If we're pinning minimum versions in fixtures, we should also state a minimum version for any soft dependencies in the README

juniorsysadmin:
I'd rather we didn't pin at all in fixtures

petems:
Ok, withdrawn 👍

PR #603, opened 2016-12-19
Changes RedHat repo to use the EPEL module

petems:
Installing the RPM is clunky and 99% of people use the EPEL module

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -7,8 +7,7 @@ fixtures:
     concat:
       repo: 'git://github.com/puppetlabs/puppetlabs-concat'
       ref: '2.1.0'
-    yum:
-      repo: 'https://github.com/voxpupuli/puppet-yum.git'
-      ref: 'v0.10.0'
+    epel:
+      repo: 'https://github.com/stahnma/puppet-module-epel'
   symlinks:
     'collectd': "#{source_dir}"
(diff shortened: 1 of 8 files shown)


petems:
Need to update metadata also...

xaque208:
+1

juniorsysadmin:
I know this is simply following the pattern, but this is a soft dependency, no?

juniorsysadmin:
I don't like the use of ```require```. Can we avoid it?
Perhaps also consider ```$repo_class``` as a variable

vinzent:
>I don't like the use of ```require```. Can we avoid it?
>Perhaps also consider ```$repo_class``` as a variable

Or just make it completly optional by requiring users to include a repo subclass if they require it? Somewhat Roles/profiles style.

petems:
I was trying to match the original behaviour as much as possible, which is to default to have the upstream class enabled.

bastelfreak:
@petems can you rebase? Then rubocop should be happy again.

petems:
@bastelfreak Done! 👍

bastelfreak:
hrm damn, the recently merged cuda plugin uses the yum module as well.

petems:
@bastelfreak Ah, fixed. 

That's repeated code that could probably be DRY-ed up, but I'll leave that to a seperate PR! :

bastelfreak:
thanks!

PR #604, opened 2016-12-19
Fixes #549: Unable to define a multi-value type

### LINKED ISSUE #549: Unable to define a multi-value type
The current 'collectd::type' does not allow for multiple values on the same line.

For example, this works:
vs_threads              value:GAUGE:0:65535

However, this is not possible:
vmpage_io               in:DERIVE:0:U, out:DERIVE:0:U

Would it be possible to extend this to allow a hash or similar to be sent to allow multiple values to be set?

jkroepke:
Hi,

this pr implement #549 in a **dirty** way. It should be refactor if puppet 3 is dropped. (https://github.com/voxpupuli/plumbing/issues/21 hopefully gets voted in the near feature... )

If you reject this because inline_template is not the state of the art I rewrite it to own defines for every type definition.

### DIFF
--- a/manifests/type.pp
+++ b/manifests/type.pp
@@ -1,29 +1,35 @@
 define collectd::type (
-  $target,
-  $ds_type,
-  $ds = $title,
-  $min = 'U',
-  $max = 'U',
-  $ds_name = 'value',
+  String                                                      $target,
+  String                                                      $ds = $name,
+  # BC compatible ....
+  Optional[Enum['ABSOLUTE', 'COUNTER', 'DERIVE', 'GAUGE']]    $ds_type = undef,
+  Variant[Numeric, Enum['U']]                                 $min = 'U',
+  Variant[Numeric, Enum['U']]                                 $max = 'U',
+  String                                                      $ds_name = 'value',
+  # BC compatible .... ending
+
+  Array[Struct[{
+    min     => Variant[Numeric, Enum['U']],
+    max     => Variant[Numeric, Enum['U']],
+    ds_type => Enum['ABSOLUTE', 'COUNTER', 'DERIVE', 'GAUGE'],
+    ds_name => String,
+  }]]                                                         $types = [],
 ) {
-  validate_string($ds_name)
-  $upper_ds_type = upcase($ds_type)
-
-  validate_re($upper_ds_type,
-              ['^ABSOLUTE$', '^COUNTER$', '^DERIVE$', '^GAUGE$'])
-
-  if $min != 'U' {
-    validate_numeric($min)
+  if empty($types) {
+    $_types = [{
+      min     => $min,
+      max     => $max,
+      ds_type => $ds_type,
+      ds_name => $ds_name,
+    }]
+  } else {
+    $_types = $types
   }
 
-  if $max != 'U' {
-    validate_numeric($max)
-  }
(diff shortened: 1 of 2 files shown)


juniorsysadmin:
Full marks for the attempt, but whenever I see code that makes me want to scrub my hands thoroughly, I start wondering about wholesale refactoring. Can we completely redo this defined type to make things easier to read and easier to maintain (regardless of Puppet 4 types)?

jkroepke:
loop in puppet 3 does not really exists.  I have to use inline_templates. Or should I code a define or every type definition?

jkroepke:
@juniorsysadmin Any news about this?

bastelfreak:
Hi @jkroepke, we dropped puppet3 in master. if you want you can migrate this PR to a puppet4 only codebase.

bastelfreak:
ping @jkroepke

jkroepke:
@bastelfreak Looks like I forgot to send my question...

What about epp templates? Can I use them instead erb files?

bastelfreak:
yep sure!

jkroepke:
Fine. Then I will start it.

jkroepke:
Okay, I have refactor this type. I'm using puppter 4 validations now.

Additions Info:
* I do a BC breaking change here:
  `ds_type` must be uppercase now. Otherwise, puppet throw an error.

jkroepke:
Additionally, the new travis the tests confused me. While the spec tests running file, the Coverage tests are failed. Its related to my PR?

bastelfreak:
@jkroepke we now run the tests in parallel, so the output might be a bit confusing. The original error is here: https://travis-ci.org/voxpupuli/puppet-collectd/jobs/200868572#L1360-L1366
and not related to coverage reports.

jkroepke:
Thanks. I squashed my commits and add a spec test for define multiple types.

manifests/type.pp
```diff
@@ -1,29 +1,35 @@
...
-  }
-
-  if $max != 'U' {
-    validate_numeric($max)
+  Array[Struct[{
+    min     => Variant[Numeric, Enum['U']],
+    max     => Variant[Numeric, Enum['U']],
+    ds_type => Enum['ABSOLUTE', 'COUNTER', 'DERIVE', 'GAUGE'],
+    ds_name => String,
+  }]]                                                         $types = [],
```
bastelfreak: I think in this line is a bit too much whitespace?

bastelfreak:
@jkroepke  thanks for the work! If I understand this correctly it is still backwardscompatible (then I can remove the label)? Plese check the one line with the whitespace, afterwards I can merge + do a release.

jkroepke:
@bastelfreak technically no, because ds_type must be an uppercase now.

In 5.x release lowercase values are accepted, too. (And transfor to uppercase)

manifests/type.pp
jkroepke: @bastelfreak I'm missing a guideline for the parameter layout w/ puppet data type definition. The reason for the much whitespace is to have all parameter variables in one vertical line. But I don't know, if it's fine.

bastelfreak: ah, now I understand the layout. I'm not aware of any current best practices for the layout, we will keep this as is.

Issue #527, opened 2016-07-05
collectd_tcp_network_connect not set to "on"


sblatt:
Hello,
Probably I am missing something major, but I'll ask anyway:
This module does not set the seboolean for collectd_tcp_network_connect, and there is no switch for it.
I already tried setting puppetagent_manage_all_files to "on" and run puppet agent again, with no success.
### Affected Puppet, Ruby, OS and module versions/distributions
CentOS Linux release 7.2.1511
puppet-collectd (v5.0.0)
ruby 2.0.0p598
puppet 3.8.7
foreman 1.11.3

Grafana 3.0.4
python-carbon 0.9.15

### What are you seeing
Collectd cannot send stats as it is not allowed to connect to the network. the error message is:
Filter subsystem: Built-in target `write': Dispatching value to all write plugins failed with status -1.
write_graphite plugin: Connecting to *grafanaserver*:2003 via tcp failed. The last error was: failed to connect to remote host: Permission denied

SELinux config:
collectd_tcp_network_connect --> off

### What behaviour did you expect instead
Data to arrive at the carbon server

### How did this behaviour get triggered
Right after installation trough puppet

### Output log
collectd: Stopping 5 write threads.
Initialization complete, entering read-loop.
write_graphite plugin: Connecting to *grafanaserver*:2003 via tcp failed. The last error was: failed to connect to remote host: Permission denied
Filter subsystem: Built-in target `write': Dispatching value to all write plugins failed with status -1.

### Any additional information you'd like to impart
Thank you for your help!

bastelfreak:
Thanks for reporting this. Are you able to create a PR with a fix as well?

sblatt:
Actually I just wanted to know if this is a bug or if I missed something. If it is indeed a bug I will happily try to fix it.

bastelfreak:
yeah, we currently don't deal with selinux at all in this module :(

sblatt:
I never worked on a puppet module, but I'll try.

bastelfreak:
If you need any help please join our IRC channel #voxpupuli, we're 24/7 happy to help.

sblatt:
I commited a change to https://github.com/sblatt/puppet-collectd/tree/selinux_support
- The variable is global to this module (as multiple plugins might need it, so they do not override eachother)
- Default is false (which translates to off, but i just like bools better)
- collectd gets restarted, which in my experience is needed for this bool to work (in both ways: on and off)
[...]

PR #607, opened 2016-12-20
Fixes for selinux

sblatt:
#527


manifests/config.pp
```diff
@@ -69,7 +73,42 @@
     purge   => $purge,
     recurse => $recurse,
   }
+ 
+  if $collectd_selinux and $::selinux == 'true' {
```
vinzent: ::selinux is a boolean and I'd expect comparing to a string will fail on Puppet4.

manifests/config.pp
```diff
@@ -69,7 +73,42 @@
...
+    selboolean { 'collectd_tcp_network_connect':
+      persistent => true,
+      provider   => getsetsebool,
+      name       => collectd_tcp_network_connect,
+      value      => 'on',
+    }
+    file { 'collectd_udev.pp':
```
vinzent: A binary is not portable across distros and not even across distro minor versions. A module compiled on EL 7.3 doest necessarily work on EL7.1 and fore sure wont work on EL6 or on Ubuntu.

sblatt:
Thynk you for your feedback. We updated to Puppet 4, so I can test that. I will also look into the binary selinux file.

oranenj:
I'm not sure if this belongs in the collectd module at all... At most it should be a completely optional manifest, and I'd not even add a dependency on puppet-selinux to metadata.json (when can we get soft dependencies?:()

Also, maybe $manage_selinux instead of $collectd_selinux?

vinzent:
I'm closing this because of no feedback of author since january.

PR #608, opened 2016-12-20
Fix python_dir fact for RedHat.

thlapin:
Use the correct method (value) for retrieving osfamily fact.
Add test for RedHat to confirm conditional is working correctly.


xaque208:
This looks good.  Thanks for the fix!

PR #617, opened 2017-01-02
BREAKING: Fix and simplify write_network plugin definition

carroarmato0:
The plugin expects a hash to be passed, yet, only the first element is being used.

Means that a lot of redundancy can be removed.

Also fixed an issue with passing the wrong value to ::collectd::plugin::network which expects servers instead of server.

Signed-off-by: Christophe Vanlancker <christophe.vanlancker@inuits.eu>

### DIFF
--- a/manifests/plugin/write_network.pp
+++ b/manifests/plugin/write_network.pp
@@ -1,23 +1,22 @@
 # A define to make a generic network output for collectd
 class collectd::plugin::write_network (
   $ensure  = 'present',
-  $servers = { 'localhost'  =>  { 'serverport' => '25826' } },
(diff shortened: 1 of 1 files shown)


carroarmato0:
Hmm, looks like the Rubocop test is complaining. But not sure if my changes have influence on that.  I just let it be?

bastelfreak:
Hi @carroarmato0 , thanks for the PR. Can you please rebase it against latest master?

bastelfreak:
Hi @carroarmato0, I'm going to close this PR due to inactivity. Feel free to reopen it if you're still interested in it.

PR #618, opened 2017-01-04
add new parameter module_import in plugin::python::module

contargo-development:
if a python-plugin sets its own name which is different from the Import Name like collectd-haproxy (https://pypi.python.org/pypi/collectd-haproxy) does:
'Import "collectd_haproxy" ' and '<Module haproxy>'
There has to be a way to be able to adjust one of these…

* set it to $module per default
* used in template for the case the name= is different from Import


bastelfreak:
Hi @contargo-development, thanks for the PR. We're currently planning another PR which fixes the rubocop issues. Please rebase after that one got merged.

zivis:
Hi @bastelfreak,
which PR is it, so I can subscribe on it, or has it already been merged?

bastelfreak:
Ah stupid me, sorry for not linking it. However it got already merged, thanks for rebasing.

zivis:
@bastelfreak no problem, thanks for merging :)

PR #622, opened 2017-01-10
Added base threshold plugin

jamtur01:
tl;dr - will try to add a template for thresholds at some point too.


rnelson0:
@jamtur01 can you add a spec test for this new type? Thanks!

jamtur01:
@rnelson0 done.

PR #625, opened 2017-01-11
Tag collectd's concat resources

### LINKED ISSUE #624: collectd adds dependency to all concat resources
- Puppet: 4.8.1
- Ruby: 2.1.5p273
- Distribution: Debian Jessie
- Module version: Master (6e767869852939ea68ff1357e56bf796d769fa03)

include collectd and another module relying on concat

## What are you seeing
If collectd packages are failing, all concat resources are not realised.
I think this is because of line [manifests/config.pp#L73](https://github.com/voxpupuli/puppet-collectd/blob/500ece66fdecf4e7778903d20fb7f93fa8d7d9c7/manifests/config.pp#L73). It seems to build a dependency for all concats, even if it should only be for concats in this module, e.g.
```
File['collectd.d'] -> Concat <| tag == 'collectd' |>
```

collectd resources are failing, everything else is working as expected.

## Output log
```
Notice: /Stage[main]/Ssh::Server::Config/Concat[/etc/ssh/sshd_config]/File[/opt/puppetlabs/puppet/cache/concat/_etc_ssh_sshd_config]: Dependency Package[collectd] has failures: true
Notice: /Stage[main]/Ssh::Server::Config/Concat[/etc/ssh/sshd_config]/File[/opt/puppetlabs/puppet/cache/concat/_etc_ssh_sshd_config]: Dependency Package[collectd-core] has failures: true
Warning: /Stage[main]/Ssh::Server::Config/Concat[/etc/ssh/sshd_config]/File[/opt/puppetlabs/puppet/cache/concat/_etc_ssh_sshd_config]: Skipping because of failed dependencies
```

rnelson0:
Fixes #624

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -70,6 +70,6 @@ class collectd::config (
     recurse => $recurse,
   }
 
-  File['collectd.d'] -> Concat <| |>
+  File['collectd.d'] -> Concat <| tag == 'collectd' |>
 
 }


rnelson0:
I'm not sure if there's a suitable spec test to use to test the ordering here, or maybe @saz can you test it for us?

saz:
I don't think this is working as expected, because this tag should be set on all concat resources in this module. Or is there some magic which adds a tag and I'm just missing the knowledge? :-)

rnelson0:
Puppet should magically add tags for every class/subclass a resource is in
automatically (collectd, collectd::sub if declared in collectd::sub).
Should :)
-- 
Rob Nelson

saz:
Cool, didn't know that. I'll try it tomorrow and let you know.

saz:
Working as expected. Thanks!

rnelson0:
You're welcome and thanks for testing!

PR #627, opened 2017-01-13
release 5.3.0

bastelfreak:


bastelfreak:
IMO we can ignore the errors, they are caused by the yum dependency which just turned puppet4 only.

dhollinger:
LGTM

PR #631, opened 2017-01-19
Stop specifying $name as the default host value

jamtur01:
This should not be the default, it doesn't make sense for localhost
bound JMX connections.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -302,7 +302,7 @@ Style/EmptyLiteral:
 Metrics/LineLength:
   Enabled: False
 
-Style/MethodCallParentheses:
+Style/MethodCallWithoutArgsParentheses:
   Enabled: True
 
 Style/MethodDefParentheses:
--- a/manifests/plugin/genericjmx/connection.pp
+++ b/manifests/plugin/genericjmx/connection.pp
@@ -2,7 +2,7 @@
 define collectd::plugin::genericjmx::connection (
   $collect,
   $service_url,
-  $host            = $name,
+  $host            = undef,
   $user            = undef,
   $password        = undef,
   $instance_prefix = undef,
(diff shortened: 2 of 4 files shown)


spec/spec_helper.rb
```diff
@@ -24,8 +24,8 @@
     puppetversion: Puppet.version,
     facterversion: Facter.version
   }
-  default_facts.merge!(YAML.load(File.read(File.expand_path('../default_facts.yml', __FILE__)))) if File.exist?(File.expand_path('../default_facts.yml', __FILE__))
-  default_facts.merge!(YAML.load(File.read(File.expand_path('../default_module_facts.yml', __FILE__)))) if File.exist?(File.expand_path('../default_module_facts.yml', __FILE__))
+  default_facts.merge!(YAML.safe_load(File.read(File.expand_path('../default_facts.yml', __FILE__)))) if File.exist?(File.expand_path('../default_facts.yml', __FILE__))
```
rnelson0: @jamtur01 hrm, we just reverted the safe_load calls earlier today. I think this may need a modulesync in the morning, and this change dropped, before we merge this.

jamtur01: I pulled master before I created this - that update wasn't in time - given the length of the time it takes the run the tests it's kinda annoying this isn't sync'ed. :(

jamtur01:
The test failure here is unrelated to the branch.

rnelson0:
It is annoying, and I'm sorry for that! We are trying to do better with the modulesyncs. Thank you for your patience.

This does look good to me and we can merge it once the other tests all pass, then I can do the modulesync in the morning to fix the build status. Thanks!

rnelson0:
And it completed while I was typing. Boom!

rnelson0:
@jamtur01 as an aside, if there is something we can do to speed up the tests, please open an issue and let us know!

PR #637, opened 2017-01-23
Parametrize destination of python config

Pigueiras:
It would very helpful for us to have a custom parameter to put a custom location of the python module configuration.


Pigueiras:
I added a test + docs explaining the new param @juniorsysadmin

xaque208:
+1

Issue #639, opened 2017-01-28
Remove quotes from the Forward option


mirekys:
Hello,

i would like to suggest removing quotes from:
https://github.com/voxpupuli/puppet-collectd/blob/master/templates/plugin/network.conf.erb#L9

Collectd (at least version 5.6.0)  is complaining
on startup when forwarding is configured using this module.

## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 4.8.2
- Ruby: 2.1.9
- Distribution: CentOS 7
- Module version: 5.3.0

## How to reproduce (e.g Puppet code you use)
```
class { 'collectd::plugin::network':
                forward => true,
```
## What are you seeing
```
#10-network.conf
<Plugin network>
    Forward "true"

</Plugin>
```
```
# startup log
collectd[]: plugin_load: plugin "network" successfully loaded.
collectd[]: cf_util_get_boolean: The Forward option requires exactly one boolean argument.
```

bastelfreak:
Hi @mirekys, thanks for the bug report. Are you able to provide a fix as well? Do you know if earlier versions of collectd also complain or only 5.6.0 and newer?

mirekys:
I cannot test on older or newer versions but by looking into the source code, I found
a somehow hidden deprecation notice, that quoted booleans will no longer be supported by collectd:
https://github.com/collectd/collectd/blob/collectd-5.6/src/daemon/configfile.c#L1135

But booleans without quotes should always work as expected, I can surely do a pull request for this.

juniorsysadmin:
Hopefully most if not all cases have been fixed by #640

PR #640, opened 2017-01-28
Remove quotes from boolean options

mirekys:
Fixes the #639 issue (broken quoted boolean handling by newer collectd versions)


alexjfisher:
@mirekys Thanks!

PR #641, opened 2017-01-29
release 6.0.0

bastelfreak:
This CHANGELOG.md was generated by the github changelog generator. Can someone please review it?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,581 +1,742 @@
-# Changelog
-
-# 2017-01-13 - Release 5.3.0
-
-This is the last release with Puppet 3 support!
-* Include custom rabbitmq types
-* Enable acceptance tests
-* Add Collectd CI Package repos
-* Add write_prometheus plugin (#600)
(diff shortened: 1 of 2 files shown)


alexjfisher:
I'd prefer the historic changelog entries were preserved.  https://github.com/skywinder/github-changelog-generator#migrating-from-a-manual-changelog

bastelfreak:
Hi. Do we really want to keep the old, hand written entries if the generator can create them as well? This brings us a unified format with proper links to all the issues and PRs.

I agree with keeping the handcrafted entries if the generator fails, this happens sometimes for past releases (like the 0.0.1 in this case).

bastelfreak:
hold of merging until https://github.com/voxpupuli/puppet-collectd/pull/604 is merged as well.

alexjfisher:
In this module, the old changelog has been very well maintained with 'bugs', 'breaking changes', 'enhancements' etc all carefully documented.  IMO Replacing this with autogenerated noise would be a travesty.

PR #643, opened 2017-02-02
Oracle Database definition must be at bottom

jkroepke:
Hi,

today, we test the module against a real oracle database and we figure out, the database configuration must be at the bottom.

### DIFF
--- a/manifests/plugin/oracle/database.pp
+++ b/manifests/plugin/oracle/database.pp
@@ -17,7 +17,7 @@ define collectd::plugin::oracle::database (
   validate_array($query)
 
   concat::fragment { "collectd_plugin_oracle_database_${name}":
-    order   => '10',
+    order   => '20',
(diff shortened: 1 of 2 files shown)


manifests/plugin/oracle/database.pp
```diff
@@ -17,7 +17,7 @@
   validate_array($query)
 
   concat::fragment { "collectd_plugin_oracle_database_${name}":
-    order   => '10',
+    order   => '20',
```
vinzent: What error message did you get  when it was not ordered at the bottom?

jkroepke: Collectd sends errors to syslog on every collect. I havn't the full messages here.

jkroepke:
@vinzent Spec Tests updated.

jkroepke:
@vinzent @bastelfreak It's possible to add this PR to the #641 ?

Without this PR, the oracle module is non functional.

PR #645, opened 2017-02-09
Added cuda GPU plugin

Yuav:

### DIFF
--- a/manifests/plugin/cuda.pp
+++ b/manifests/plugin/cuda.pp
@@ -12,11 +12,6 @@
 #   Passed to package and collectd::plugin resources ( both )
 #   Default: present
 #
-# [*interval*]
-#   Integer
-#   Interval setting for the plugin
-#   Default: undef
-#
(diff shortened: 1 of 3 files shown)


Yuav:
The acceptance test for Debian timed out. What's the preferred way to trigger restart of that job?

manifests/plugin/cuda.pp
```diff
@@ -0,0 +1,78 @@
+# cuda plugin
```
vinzent: I would start directly with puppet-strings docs : https://github.com/puppetlabs/puppet-strings#documenting-puppet-classes-and-defined-types

bastelfreak: +1 for puppet-strings

Yuav: Got an example?

Yuav:
This what you're looking for?

bastelfreak:
Thanks @Yuav!

PR #646, opened 2017-02-09
df: add devices parameter

maage:
df plugin has had Device parameter about 10 years,
https://collectd.org/documentation/manpages/collectd.conf.5.shtml#plugin_df
https://collectd.org/wiki/index.php/Plugin:DF
add support for it as devices parameter


bastelfreak:
Thanks @maage!

PR #655, opened 2017-03-10
modulesync 0.20.1

bastelfreak:

### DIFF
--- a/spec/acceptance/nodesets/docker/ubuntu-12.04.yml
+++ b/spec/acceptance/nodesets/docker/ubuntu-12.04.yml
@@ -12,8 +12,6 @@ HOSTS:
     docker_image_commands:
       - 'apt-get install -y net-tools wget'
       - 'locale-gen en_US.UTF-8'
-      - 'ln -s /opt/puppetlabs/bin/facter /usr/bin/facter'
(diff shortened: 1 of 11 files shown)


spec/acceptance/nodesets/docker/ubuntu-12.04.yml
```diff
@@ -12,8 +12,6 @@ HOSTS:
     docker_image_commands:
       - 'apt-get install -y net-tools wget'
       - 'locale-gen en_US.UTF-8'
-      - 'ln -s /opt/puppetlabs/bin/facter /usr/bin/facter'
```
vinzent: @petems you added these 2 symlinks in the modulesync managed file with https://github.com/voxpupuli/puppet-collectd/commit/8dba7fc508696e72d8e513c7b62520960eff5a57 . is it really required?

petems: @vinzent I think depending on the version of beaker used, the path doesn't get exported and it's needed. Let me check.

bastelfreak: ping @petems

bastelfreak:
I'm going to merge this as is. We can still change it in the future if we notice that the symlinks are required.

PR #657, opened 2017-03-15
openldap - add binddn and password parameters

leonkyneur:
These parameters are required if the directory server requires authentication to retrieve statistics.


bastelfreak:
Hi @leonkyneur, thanks for the PR. are you able to add spec tests as well?

leonkyneur:
Hi @bastelfreak may need guidance on the tests -

I did look at the existing spec tests and given these params are just part of the instances hash and are optional - I couldn't see merit extending the existing tests as there is no validation on the values of 'instances'.

Do you think by extending the current tests to just include these params and checking with "contain_file().with(content:" suffice? 

Or would you prefer some way of iterating the hash and checking the parameters?

oranenj:
@leonkyneur it will be enough to just extend the params hash and checking the generated file, yes.

I see the ldap spec isn't very easy to read, but that shouldn't be too difficult to modify. If you feel up to it, a re-format of the spec using eg. heredoc syntax might make it easier to extend.

leonkyneur:
agree with you @oranenj re heredoc - I've modified tests to utilize this. 

Latest commit includes the new params in the spec test.

leonkyneur:
@oranenj commit squashed. Thanks

oranenj:
And merged. Thanks!

PR #659, opened 2017-03-29
add $measureresponsecode for curl plugin

azhurbilo:
according this documentation page https://collectd.org/documentation/manpages/collectd.conf.5.shtml#plugin_curl

we can start collecting response code

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -49,11 +49,11 @@ class collectd (
   anchor { 'collectd::begin': }
   anchor { 'collectd::end': }
 
-  Class['::collectd::repo'] ~>
-  Class['::collectd::install']
+  Class['::collectd::repo']
+  ~> Class['::collectd::install']
 
-  Anchor['collectd::begin'] ->
-  Class['collectd::install'] ->
-  Class['collectd::config'] ~>
-  Class['collectd::service']
+  Anchor['collectd::begin']
+  -> Class['collectd::install']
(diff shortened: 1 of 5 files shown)


azhurbilo:
Yep, I also saw this failed check before starting add any changes, so it's not related to my MR )

```
Warning: Dependency puppetlabs/stdlib has an open ended dependency version requirement >= 4.6.0
Warning: Dependency puppetlabs/concat has an open ended dependency version requirement >= 1.2.5
Warning: Dependency stahnma-epel has an open ended dependency version requirement >= 1.2.2
Warning: Dependency puppetlabs-apt has an open ended dependency version requirement >= 2.2.0
manifests/init.pp:52:arrow_on_right_operand_line:WARNING:arrow should be on the right operand's line
manifests/init.pp:55:arrow_on_right_operand_line:WARNING:arrow should be on the right operand's line
manifests/init.pp:56:arrow_on_right_operand_line:WARNING:arrow should be on the right operand's line
manifests/init.pp:57:arrow_on_right_operand_line:WARNING:arrow should be on the right operand's line
```

bastelfreak:
The 4 warnings about dependencies can be ignoried. We always test against master branch, not releases, to catch errors before they are released.

The other warnings are caused by a lately updated puppet-lint plugin. Are you able to fix them? Otherwise we can do it in a separate pull request.

oranenj:
Looks good to me. Thanks :)

PR #660, opened 2017-03-30
Update `collectd::plugin::python::module` README

sigv:
The module definitions now need arrays for configurations. README was updated to show that for defining multiple modules, but the single-module definition was forgotten.


sigv:
`rake test` seems to be failing the manifests - fixing that is not really related to the scope of this PR.

oranenj:
Yeah, and I'm pretty sure the tests are passing now anyway. There's just no way to re-run them. I'll just merge this. Thanks

PR #667, opened 2017-05-09
Change 'Redhat' to 'RedHat' when comparing against the osfamily fact

bodgit:
I was tracking down why we had forked our own copy of this module and found one of the changes was simply changing the case of `Redhat` to `RedHat` in some common places. Before I remembered that Puppet helpfully compares strings case insensitively I updated all occurrences given the module uses both forms throughout. Should Puppet ever change its behaviour the latter form matches exactly what facter returns.

To keep my employers legal team happy, I have to include the following:
> This contribution is provided 'as is' and without any warranty or guarantee of any kind, express or implied, including in relation to its quality, suitability for a particular purpose or non-infringement. To the extent permitted by law, in no event shall the creator of this contribution be liable for any claim, damage or other liability, whether arising in contract, tort or otherwise, arising out of or in connection with this contribution.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -56,7 +56,7 @@ class collectd::params {
       $default_python_dir = '/opt/csw/share/collectd/python'
       $manage_repo        = false
     }
-    'Redhat': {
+    'RedHat': {
       $package_name       = 'collectd'
       $package_provider   = 'yum'
       $collectd_dir       = '/etc/collectd.d'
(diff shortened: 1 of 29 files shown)


oranenj:
The acceptance tests failing doesn't seem to be because of these changes... Hmm.

oranenj:
I'll try re-running them.

oranenj:
I changed the title to be a bit more descriptive as it may appear in the changelog. Thanks!

bodgit:
Yes, this and my other two PR's seem to have failed with the same unrelated error.

PR #671, opened 2017-05-25
Install 'collectd-python' package on Amazon Linux.

irgeek:
When the package installation check in `manifests/plugin/python.pp` changed from all RedHat distress to Fedora the module stopped working on Amazon Linux. This change updates the check to install the module on Amazon Linux.

### DIFF
--- a/manifests/plugin/python.pp
+++ b/manifests/plugin/python.pp
@@ -38,7 +38,7 @@ class collectd::plugin::python (
     $ensure_real = 'absent'
   }
 
-  if $::operatingsystem == 'Fedora' {
+  if $facts['os']['name'] == 'Fedora' or $facts['os']['name'] == 'Amazon' {
     if $_manage_package {
       package { 'collectd-python':
         ensure => $ensure_real,


irgeek:
It looks like tests are failing because of (very) recent changes to the puppetlabs-apt module -- it's now referencing `$facts['os']['name']`.

manifests/plugin/python.pp
```diff
@@ -38,7 +38,7 @@
     $ensure_real = 'absent'
   }
 
-  if $::operatingsystem == 'Fedora' {
+  if $::operatingsystem == 'Fedora' or $::operatingsystem == 'Amazon' {
```
bastelfreak: can you use the new facts syntax here? $facts['os']['name'] is probably correct.

spec/default_module_facts.yml
```diff
@@ -4,3 +4,7 @@ python_dir: '/usr/local/lib/python2.7/dist-packages'
 operatingsystem: 'Debian'
 lsbdistid: 'Debian'
 lsbdistcodename: 'wheezy'
+os:
```
bastelfreak: This is, in my opinion, a bad idea. We should instead migrate the failing tests to rspec-puppet-facts. I will have a look at it and provide a PR

bastelfreak:
Hi @irgeek , thanks for the PR. Can you please rebase it against latest master?

bastelfreak:
@irgeek ping

irgeek:
@bastelfreak I've rebased this and updated to use `$facts['os']['name']`, but tests on master are failing at the moment so they're failing here too.

bastelfreak:
@irgeek we fixed this in master. Thanks for the PR.

PR #672, opened 2017-05-26
Modulesync 0.21.3

bastelfreak:

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -6,7 +6,6 @@ fixtures:
       repo: 'git://github.com/puppetlabs/puppetlabs-stdlib'
     concat:
       repo: 'git://github.com/puppetlabs/puppetlabs-concat'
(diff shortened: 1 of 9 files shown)


bastelfreak:
I only wanted to do the rspec-puppet-facts integration, but tests would take way too long without the parallel_tests gem, so I included a modulesync.

bastelfreak:
blargh

bastelfreak:
I pulled most of the changes out into another branch, to make the diff smaller. They are now at https://github.com/voxpupuli/puppet-collectd/pull/673. I don't think that we can merge this if we want to review all changes properly + have travis green. So we should merge it in parts.

bastelfreak:
I think this is safe to merge. all the acceptance tests are still green, just unit tests fail because of issues in the puppetlabs-apt modules. This will be addressed in another PR.

PR #673, opened 2017-05-27
rspec-puppet-facts integration

bastelfreak:


bastelfreak:
this got pulled out from https://github.com/voxpupuli/puppet-collectd/pull/672

bastelfreak:
spec tests itself are fine, but they are failling because they take too much time. I'm not sure how we should deal with this. We can probably drop a huge amount of existing tests.

tpoindessous:
Hi

the test are failing because of Travis ? If so, can this page help you ? 

[https://docs.travis-ci.com/user/common-build-problems/#My-builds-are-timing-out](https://docs.travis-ci.com/user/common-build-problems/#My-builds-are-timing-out)

They said that you can prefix your command with `travis_wait` like this :
`
install: travis_wait mvn install
`

If it's not that, sorry to bother you.

Thanks !

alexjfisher:
@bastelfreak Instead of dropping tests, you could just test a subset of the OSes instead of *all* the ones from metadata.json.

ekohl:
In the foreman we do this through a spec helper: https://github.com/theforeman/foreman-installer-modulesync/blob/cce7db684851ce28aad8c5af476cd77c81dbdbdf/moduleroot/spec/spec_helper.rb.erb#L21-L45

PR #674, opened 2017-05-27
drop legacy OS from support matrix

bastelfreak:
we should not support outdated operating systems. They all went EOL in
the past months. This also reduces our test-runtime significatly.


bastelfreak:
this got pulled out from https://github.com/voxpupuli/puppet-collectd/pull/673

PR #677, opened 2017-06-06
Fix ping plugin interval mixup

maage:
After collectd 5.2 ping plugin interval was also defined as plugin interval. This patch adds ping_interval to be interval between pings and leaves interval as plugin level interval (how oftern to collect stats).

Other Interval is defined at templates/loadplugin.conf.erb.

This patch tries to be compatible with collecd version before 5.2, but when you do update to 5.2 or newer you need to define ping_interval for ping plugin to work as expected. If support for collectd before 5.2 is dropped, logic can be simplified.

Other cosmetic patch just drops extra newlines from created configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1241,7 +1241,9 @@ collectd::plugin::perl::plugin {
 ```puppet
 collectd::plugin::ping {
   'example':
-    hosts => ['example.com'],
+    hosts         => ['example.com'],
+    ping_interval => '0.100',
+    timeout       => '0.095',
 }
 ```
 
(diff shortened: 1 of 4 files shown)


bastelfreak:
Hi @maage , thanks for the PR. Can you please rebase it against latest master?

oranenj:
Fixed the merge conflicts here too. I think.

wyardley:
This has failing tests and also some rubocop warnings to clear up.

bastelfreak:
can you please rebase? We fixed the spec issues in our master branch.

bastelfreak:
Hi @maage, I'm going to close this PR due to inactivity. Please feel free to reopen if you're still interested in it.

PR #621, opened 2017-01-06
Add support for nut plugin and fix os specific location of typesdb

amishHammer:
Add support for nut plugin and fix os specific location of typesdb.

PR #681, opened 2017-06-13
Add support for nut plugin (#621)

trustchk:
This is a rebase of @amishHammer PR #621 against current master but without the change to the types.db handling and with a fix for the linter warning. It also includes a spec test (which might need some additions).


oranenj:
It looks like the spec overhaul caused this to break, and need a rebase/fixup again.

bastelfreak:
@trustchk can you please rebase? We fixed the spec issues in our master branch.

PR #682, opened 2017-06-14
add fscache plugin

wildente:
https://collectd.org/wiki/index.php/Plugin:FSCache


oranenj:
checks failed, but only because travis is silly. I'm merging this anyway, since it's pretty trivial

Thanks for the contribution :)

PR #685, opened 2017-06-29
Remove interval parameter from write plugins.

kbor:
Hello, to my knowledge it makes no sense to configure interval for write plugins. Collectd has no concept of interval for write plugins and if configured, it is silently ignored. It is therefore better to remove it from the configuration so that manifests don't give a false impression that this parameter does something useful.

### DIFF
--- a/manifests/plugin/write_graphite.pp
+++ b/manifests/plugin/write_graphite.pp
@@ -2,7 +2,6 @@
 class collectd::plugin::write_graphite (
   $carbons           = {},
   $carbon_defaults   = {},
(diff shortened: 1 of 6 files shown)


oranenj:
There seem to be some puppet lint issues. Could you fix those?

bastelfreak:
Hi @kbor, thanks for the PR. Can you please rebase it against latest master?

kbor:
Hi, yes, sorry, didn't notice. Lint fixed and rebased.

bastelfreak:
Thanks!

PR #692, opened 2017-09-06
bump version on puppetlabs/apt dependency

costela:
This seems to require no changes on collectd's side.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -104,7 +104,7 @@
     },
     {
       "name": "puppetlabs-apt",
-      "version_requirement": ">= 2.2.0 < 3.0.0"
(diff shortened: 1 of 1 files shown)


bastelfreak:
Hi @costela, thansk for the PR! The commit message is a bit missleading. Why do you change the concat dependency as well?

costela:
@bastelfreak oops, that slipped in. Should be clean now.

wyardley:
btw, I think apt is supposed to be a soft dependency per https://docs.puppet.com/puppet/5.1/style_guide.html#dependencies

costela:
@wyardley good point.

Issue #693, opened 2017-09-15
IPTables plugin doesn't support IPv6


herver:
## Affected Puppet, Ruby, OS and module versions/distributions
- Puppet: 4.9.4
- Ruby: 2.0.0p648
- Distribution: Centos7
- Module version: 5.3.0

## How to reproduce (e.g Puppet code you use)
I'd like to use the `Chain6` feature of [Collectd's IPTables plugin](https://collectd.org/wiki/index.php/Plugin:IPTables) but the puppet manifest doesn't support it.

PR #694, opened 2017-09-15
Add support for Chain6 directive

herver:
This PR is related to issue #693 and adds support for the `Chain6` configuration directive of the `iptables` Collectd plugin.

Closes GH-693

### DIFF
--- a/examples/plugins/iptables.pp
+++ b/examples/plugins/iptables.pp
@@ -6,7 +6,6 @@ class { '::collectd::plugin::iptables':
     'filter' => 'HTTP',
   },
   chains6 => {
-    'nat'    => 'In6_SSH',
     'filter' => 'HTTP6',
   },
 }
--- a/examples/plugins/iptables.pp
+++ b/examples/plugins/iptables.pp
@@ -1,8 +1,12 @@
 include ::collectd
 
 class { '::collectd::plugin::iptables':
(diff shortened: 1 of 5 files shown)


examples/plugins/iptables.pp
```diff
@@ -1,8 +1,12 @@
 include ::collectd
 
 class { '::collectd::plugin::iptables':
-  chains => {
+  chains  => {
     'nat'    => 'In_SSH',
     'filter' => 'HTTP',
   },
+  chains6 => {
+    'nat'    => 'In6_SSH',
```
ekohl: A nat example for IPv6 is odd. I'd just leave this line out

spec/classes/collectd_plugin_iptables_spec.rb
```diff
@@ -26,6 +26,20 @@
         end
       end
 
+      context ':ensure => present and :chains6 => { \'nat\' => \'In6_SSH\' }' do
```
ekohl: Can you also use ```filter``` instead of ```nat```? I know it doesn't matter that much, but it feels wrong :)

traylenator:
@herver the example in the README could also be updated, ... two places for examples with this module.

traylenator:
Sorry should have squashed commits first, to late now.

PR #696, opened 2017-09-30
modulesync 1.2.0

wyardley:


TraGicCode:
It looks like the build for this module has been broken for a while.  Think we can fix this?

TraGicCode:
I'm going to merge this.  It looks like a big refactoring took place and things started getting merged in before the tests were even working so it's really hard to find out what change is causing the issue.  Maybe @bastelfreak might have an idea?

PR #701, opened 2017-10-29
plugin::apache - Use data types

juniorsysadmin:
- Use data types for the Apache plugin
- Bump up the puppetlabs-stdlib version so that Stdlib::Httpurl can be
  used

### DIFF
--- a/manifests/plugin/apache/instance.pp
+++ b/manifests/plugin/apache/instance.pp
@@ -1,12 +1,12 @@
 # https://collectd.org/wiki/index.php/Plugin:Apache
 define collectd::plugin::apache::instance (
-  $url,
-  $ensure = present,
-  $user = undef,
-  $password = undef,
-  $verifypeer = undef,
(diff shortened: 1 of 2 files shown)


manifests/plugin/apache/instance.pp
```diff
@@ -1,12 +1,12 @@
...
-  $ensure = present,
-  $user = undef,
-  $password = undef,
-  $verifypeer = undef,
-  $verifyhost = undef,
-  $cacert = undef,
+  Stdlib::Httpurl $url,
+  String $ensure                         = present,
+  Optional[String] $user                 = undef,
+  Optional[String] $password             = undef,
```
juniorsysadmin: I tried to use Sensitive, but couldn't get it working with `$password` being optional

bastelfreak: small steps, still better than no datatypes \o/

juniorsysadmin:
Depends on #703

Issue #700, opened 2017-10-24
Unknown variable collectd_version


fnoop:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 5.2.0
- Ruby: 2.3.3p222
- Distribution: Ubuntu 17.04 zesty
- Module version: 6.0.0

## How to reproduce (e.g Puppet code you use)
class { "collectd":
        config_file     => $config_file,
        plugin_conf_dir => $plugin_conf_dir,
        plugin_conf_dir_mode => "755",
        purge           => true,
        recurse         => true,
        purge_config    => true,
        minimum_version => '5.7',
        manage_package  => $manage_package,
        manage_repo     => $manage_repo,
        manage_service  => true,
        service_name    => 'maverick-collectd',
        typesdb         => [$typesdb],
    }

## What are you seeing
Error: Evaluation Error: Unknown variable: '::collectd_version'. at /srv/maverick/software/maverick/manifests/puppet-modules/collectd/manifests/init.pp:37:33 on node maverick-ubuntu17

## Any additional information you'd like to impart
puppet is being run as a local standalone process (puppet apply), not as client/server.  Worked fine with version 5.3.0 of this module

ctriegg:
The problem occurs when `strict_variables` are enabled in the puppetmaster configuration: https://puppet.com/docs/puppet/5.2/configuration.html#strictvariables

When the fact is empty when collectd is not installed (that's the case on the first puppet run with this module) it fails.

actatux:
Hi @fnoop ,

PR #708 merged yesterday. Can you give a try in your environment?

PR #707, opened 2017-11-03
Collectd version fact value

ctriegg:
fixes issue #700

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -34,7 +34,7 @@ class collectd (
   $write_threads                                                      = $collectd::params::write_threads,
 ) inherits collectd::params {
 
-  $collectd_version_real = pick($::collectd_version, $minimum_version)
+  $collectd_version_real = pick_default($facts['collectd_version'], $minimum_version)
 
   class { '::collectd::install':
     package_install_options => $package_install_options,


bastelfreak:
Strange thing here is that our tests were green before this change, so they didn't detect this error. Conclusion: We don't have a test that proves that this change fixes anything / breaks nothing. In theory the puppetlabs_spec_helper enables STRICT_VARIABLES. It needs to be checked if this is actually the case, or if we need more tests here.

bastelfreak:
@ctriegg cna you also please check the used email address in your commit? It isn't associated with your github account.

ekohl:
@bastelfreak we ensure the fact is present in our tests: https://github.com/voxpupuli/puppet-collectd/blob/15a028feb63f64444ce837c3f7397c46d219d13e/spec/default_module_facts.yml#L2

alexjfisher:
I'm not sure this is the right fix.  As per our own docs, we shouldn't be using [top scope variables to access facts](https://github.com/voxpupuli/voxpupuli.github.io/blame/master/docs/index.md#L125).

I think changing [init.pp](https://github.com/voxpupuli/puppet-collectd/blob/15a028feb63f64444ce837c3f7397c46d219d13e/manifests/init.pp#L37) line 37
```
collectd_version_real = pick($::collectd_version, $minimum_version)
```
to
```
collectd_version_real = pick_default($facts['collectd_version'], $minimum_version)
```
would be better.

alexjfisher:
Actually, I think this isn't just a style thing.  This PR introduces a bug.
`collectd_version_real` should equal `minimum_version` if the fact isn't available, not `false`.

ctriegg:
alexjfisher is right. 
adapting `pick` to `pick_default` is the better solution. 

just running tests, pushing it when they're finished

alexjfisher:
@ctriegg I've already got https://github.com/voxpupuli/puppet-collectd/pull/708 open (which includes a similar fix for `python_dir`)

alexjfisher:
@ctriegg Thank you for looking into this.

PR #708, opened 2017-11-05
Fix strict variables error when facts are missing

alexjfisher:
An alternative to https://github.com/voxpupuli/puppet-collectd/pull/707

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -34,7 +34,7 @@ class collectd (
   $write_threads                                                      = $collectd::params::write_threads,
 ) inherits collectd::params {
 
-  $collectd_version_real = pick($::collectd_version, $minimum_version)
+  $collectd_version_real = pick_default($facts['collectd_version'], $minimum_version)
 
   class { '::collectd::install':
     package_install_options => $package_install_options,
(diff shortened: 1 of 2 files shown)


manifests/init.pp
```diff
@@ -34,7 +34,7 @@
   $write_threads                                                      = $collectd::params::write_threads,
 ) inherits collectd::params {
 
-  $collectd_version_real = pick($::collectd_version, $minimum_version)
+  $collectd_version_real = pick_default($facts['collectd_version'], $minimum_version)
```
alexjfisher: I'm not sure if `minimum_version` is allowed to be set to `undef`, but `pick_default` would allow this.

manifests/params.pp
```diff
@@ -135,5 +135,5 @@
   }
 
   # Override with custom fact value (present only if python is installed)
-  $python_dir            = pick($::python_dir, $default_python_dir)
+  $python_dir = pick($facts['python_dir'], $default_python_dir)
```
alexjfisher: `default_python_dir` is always set, so `pick_default` isn't needed/wanted here.

Issue #652, opened 2017-02-22
Process plugin silently breaks processmatch/process defines.


bzed:
@devfaz with 395036ae332c56800481459fef9f1691a484de9f you've **silently** broken the usage of collectd::plugin::processes::processmatch and collectd::plugin::processes::process as the concat fragments from both defines are not collected anymore unless you've specified a process or processmatch option in the collectd::plugin::processes class.

The correct fix would be to use collectd from wheezy-backports or maybe make collectd monitor itself by default.

juniorsysadmin:
Fixed by #712

Issue #595, opened 2016-12-10
Processes plugin creates strange config


gytisgreitai:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 3.8.7
- Ruby: 1.9.3p484
- Distribution: Ubuntu 14.04
- Module version: 5.1

## How to reproduce (e.g Puppet code you use)
Code taken from README.md:
```
  collectd::plugin::processes::process { 'collectd' : }
  collectd::plugin::processes::processmatch { 'elasticsearch' :
    regex => '.*java.*org.elasticsearch.bootstrap.Elasticsearch'
  }
```
## What are you seeing
It generates the following once run on the server:
```
10-processes.conf:
<LoadPlugin processes>
  Globals false
</LoadPlugin>
```

## What behaviour did you expect instead
I would expect it to contain definitions for elastic and collectd service.

If i use 
   ```
 class { 'collectd::plugin::processes':
      processes => ['process1', 'process2'],
      process_matches => [
        { name => 'process-all', regex => 'process.*' }
      ],
    }
```
everything works as expected

## Output log

> Info: Concat[/etc/collectd/conf.d/processes-config.conf]: Scheduling refresh of Service[collectd]
> Info: Class[Collectd::Config]: Scheduling refresh of Class[Collectd::Service]
> Info: Class[Collectd::Service]: Scheduling refresh of Service[collectd]

juniorsysadmin:
It *looks* like https://github.com/voxpupuli/puppet-collectd/blob/master/examples/plugins/processes.pp is correct, README is wrong.

oscarkraemer:
This seems to be the same issue as #652

juniorsysadmin:
Fixed by #712

PR #712, opened 2017-11-09
Make it possible to add processes one by one

oscarkraemer:
This fixes a old silent bug that was created by
395036ae332c56800481459fef9f1691a484de9f

This commit makes it possible define single process again like:
collectd::plugin::processes::process { 'collectd' : }
Without also having to create:
class { 'collectd::plugin::processes':
  processes => [ 'foo', 'bar' }

This fixes the issues #652 and #595.

### DIFF
--- a/manifests/plugin/processes.pp
+++ b/manifests/plugin/processes.pp
@@ -15,14 +15,8 @@ class collectd::plugin::processes (
     interval => $interval,
   }
 
-  if ( $processes or $process_matches ) {
-    $process_config_ensure = 'present'
-  } else {
-    $process_config_ensure = absent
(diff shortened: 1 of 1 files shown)


oscarkraemer:
It would have been possible to create the same affect as 395036a tried to solve by calling the class with ensure => absent .

juniorsysadmin:
@oscarkraemer (1) Would it be possible to add test cases to prove that this fixes the issues listed (2) Does the README need to be updated?

oscarkraemer:
I will not write any new tests, I don't have time. The README should not need any updates since these changes would make the readme correct.

PR #717, opened 2017-11-12
collectd::plugin::exec::cmd clean up

juniorsysadmin:
- Add data types for user and group
- Remove $ensure as it is currently unused
- Remove long deprecated handling of old file path


bastelfreak:
@juniorsysadmin do we need to label this with backwards-incompatible?

juniorsysadmin:
I would say no

PR #721, opened 2017-11-12
plugin::filter::match - Use data types

juniorsysadmin:

### DIFF
--- a/manifests/plugin/filter/match.pp
+++ b/manifests/plugin/filter/match.pp
@@ -1,17 +1,13 @@
 # https://collectd.org/wiki/index.php/Chains
 define collectd::plugin::filter::match (
-  $chain,
-  $rule,
-  $plugin,
-  $options = undef,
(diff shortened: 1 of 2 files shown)


manifests/plugin/filter/match.pp
```diff
@@ -1,18 +1,15 @@
 # https://collectd.org/wiki/index.php/Chains
 define collectd::plugin::filter::match (
-  $chain,
-  $rule,
-  $plugin,
-  $options = undef,
+  String $chain,
+  String $rule,
+  # Same as listed in $collectd::plugin::filter::plugin_matches
+  Enum['regex', 'timediff', 'value', 'empty_counter', 'hashed'] $plugin,
```
bastelfreak: should we create a custom datatype for this?

juniorsysadmin: Yes

PR #731, opened 2017-11-17
added: check params variable in query pg template

vaclandic:
Sometimes needs make query without parameters, for example check lags on slaves:
"SELECT EXTRACT(EPOCH FROM (now() - pg_last_xact_replay_timestamp()))::INT"


bastelfreak:
Thanks for this PR @vaclandic!

PR #735, opened 2017-11-30
Escape json posts in curl_json plugin

alexjfisher:
*If* in the 'post' parameter we've been given a valid json string, it
definately won't work unless we do some escaping of double quotes.

If any strings in the JSON also contain (correctly escaped) double
quotes, even more backslashes are needed.

eg. The following JSON

```
{"param1":"something with a double quote \" in it"}
```

Would end up in the config file as

```
Post "{\"param1\":\"something with a double quote \\\" in it\"}"
```

### DIFF
--- a/spec/classes/collectd_plugin_fscache_spec.rb
+++ b/spec/classes/collectd_plugin_fscache_spec.rb
@@ -13,6 +13,7 @@ describe 'collectd::plugin::fscache', type: :class do
     let :facts do
       {
         osfamily: 'FreeBSD',
+        operatingsystem: 'FreeBSD',
         os: { family: 'FreeBSD' },
         collectd_version: '4.7.0',
         operatingsystemmajrelease: '7',
@@ -31,6 +32,7 @@ describe 'collectd::plugin::fscache', type: :class do
     let :facts do
       {
         osfamily: 'FreeBSD',
+        operatingsystem: 'FreeBSD',
         os: { family: 'FreeBSD' },
         collectd_version: '4.7.0',
         operatingsystemmajrelease: '7',
--- a/spec/defines/collectd_plugin_curl_json_spec.rb
+++ b/spec/defines/collectd_plugin_curl_json_spec.rb
@@ -28,8 +28,8 @@ describe 'collectd::plugin::curl_json', type: :define do
           verifypeer: 'false',
           verifyhost: 'false',
           cacert: '/path/to/ca.crt',
-          header: 'Accept: application/json',
-          post: '{secret: \"mysecret\"}',
+          header: 'Content-Type: application/x-www-form-urlencoded',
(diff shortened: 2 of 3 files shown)


templates/curl_json.conf.erb
```diff
@@ -40,10 +40,22 @@ LoadPlugin "curl_json"
 <% if @header -%>
     Header "<%= @header %>"
 <% end -%>
+<%-
+  def valid_json?(json)
```
alexjfisher: I'm not sure this is a great idea.  Suggestions welcome!

templates/curl_json.conf.erb
```diff
@@ -40,10 +40,22 @@ LoadPlugin "curl_json"
...
+    JSON.parse(json)
+    return true
+  rescue JSON::ParserError
+    return false
+  end
+-%>
 <% unless @post.nil? -%>
+<% if valid_json?(@post) -%>
+    Post "<%= @post.gsub('\\"','\\\\\\\\\"').gsub!(%r{(?<!\\)"},'\"') %>"
```
alexjfisher: I'm very very sorry!  All those backslashes are needed though.  See https://www.ruby-forum.com/topic/143645

templates/curl_json.conf.erb
```diff
@@ -40,10 +40,22 @@ LoadPlugin "curl_json"
...
+<% else -%>
     Post "<%= @post %>"
```
alexjfisher: If it's not JSON, nothing is changed. If it is, it definitely won't work if we don't do any escaping.

spec/defines/collectd_plugin_curl_json_spec.rb
```diff
@@ -28,8 +28,8 @@
           verifypeer: 'false',
           verifyhost: 'false',
           cacert: '/path/to/ca.crt',
-          header: 'Accept: application/json',
-          post: '{secret: \"mysecret\"}',
```
alexjfisher: This wasn't valid JSON.

spec/classes/collectd_plugin_fscache_spec.rb
```diff
@@ -13,6 +13,7 @@
     let :facts do
       {
         osfamily: 'FreeBSD',
+        operatingsystem: 'FreeBSD',
```
alexjfisher: Unrelated change to fix intermittent test failures.

bastelfreak: could you migrate this to rspec puppet facts?

alexjfisher: Love to, but there don't appear to be any. https://github.com/camptocamp/facterdb/tree/master/facts/2.5

alexjfisher:
@bastelfreak You happy with this?

Issue #688, opened 2017-07-25
service resource name 'collectd' should be renamed


fnoop:
## Affected Puppet, Ruby, OS and module versions/distributions
- Puppet: 5.0.0
- Ruby: 2.1.5p273
- Distribution: Raspbian Jessie
- Module version: 5.3.0

## How to reproduce (e.g Puppet code you use)
```
    class { "collectd":
        purge           => true,
        recurse         => true,
        purge_config    => true,
        minimum_version => '5.4',
        manage_service  => true,
        service_name   => 'maverick-collectd',
    } ->
    service_wrapper { "collectd":
        ensure          => stopped,
        enable          => false,
    }
```

## What are you seeing
If I set the service_name to be a custom service, I cannot then also control the collectd system service.
```
Error: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Error while evaluating a Resource Statement, Duplicate declaration: Service[collectd] is already declared in file /srv/maverick/software/maverick/manifests/puppet-modules/collectd/manifests/service.pp:9; cannot redeclare at /srv/maverick/software/maverick/manifests/init.pp:176 at /srv/maverick/software/maverick/manifests/init.pp:176:9  at /srv/maverick/software/maverick/manifests/maverick-modules/maverick_analysis/manifests/collect.pp:28 on node maverick-raspberry.home
```

## What behaviour did you expect instead
I expect to be able to control 'collectd' system service as normal.

## Any additional information you'd like to impart
This is because the name of the module service resource is set to 'collectd', so conflicting with any other use of the 'collectd' service.  It could be renamed to something like 'collectd_service' to not clash.

fnoop:
There are a lot of resources that notify the service collectd:
```
[dev] [mav@maverick-raspberry ~/software/maverick/manifests/puppet-modules/collectd/manifests]$ grep -r collectd |grep Service
plugin/write_graphite.pp:        notify         => Service['collectd'],
plugin/network/server.pp:    notify  => Service['collectd'],
plugin/network/listener.pp:    notify  => Service['collectd'],
plugin/filter/chain.pp:    notify         => Service['collectd'],
plugin/rabbitmq.pp:    notify  => Service['collectd'],
plugin/exec.pp:    notify         => Service['collectd'],
plugin/genericjmx.pp:    notify         => Service['collectd'],
plugin/filecount/directory.pp:    notify  => Service['collectd'],
[...]

PR #738, opened 2017-12-04
Use service_name param instead of static collectd, #688

fnoop:
This PR is to fix behaviour reported in #688.  puppet-collectd currently refers to the controlled service as `Service['collectd']`.  This means that if you specify $collectd::service_name, you still cannot control the actual 'collectd' service outside of puppet-collectd.  This PR changes all references from 'collectd' to $collectd::service_name.

### DIFF
--- a/manifests/plugin.pp
+++ b/manifests/plugin.pp
@@ -20,7 +20,7 @@ define collectd::plugin (
     group   => $root_group,
     mode    => '0640',
     content => template('collectd/loadplugin.conf.erb'),
-    notify  => Service['collectd'],
+    notify  => Service[$collectd::service_name],
   }
 
   # Older versions of this module didn't use the "00-" prefix.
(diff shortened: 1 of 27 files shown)


alexjfisher:
Could you look at the test failures? https://github.com/voxpupuli/puppet-collectd/blob/98e071b22bb9264f416052e4c6896dbf49f6b89b/spec/defines/collectd_plugin_apache_instance_spec.rb#L29 needs updating since the service_name doesn't default to 'collectd' for all supported OSes.

fnoop:
Is this because the manifests now refer to "service { $service_name:" instead of "service { 'collectd':"?
I have no idea how to fix the tests for this, sorry.

bastelfreak:
We implemented this in https://github.com/voxpupuli/puppet-collectd/pull/771

PR #743, opened 2017-12-15
Adds class ovs_events plugin

paramite:
This patch is adding support for ovs_events plugin configuration and installation.

### DIFF
--- a/manifests/plugin/ovs_events.pp
+++ b/manifests/plugin/ovs_events.pp
@@ -1,0 +1,70 @@
+#== Class: collectd::plugin::ovs_events
+#
+# Class to manage ovs_events plugin for collectd
+#
+# Documentation:
+#   https://collectd.org/documentation/manpages/collectd.conf.5.shtml#plugin_ovs_events
+#
+# === Parameters
+#
+# [* address *]
+#  The address of the OVS DB server JSON-RPC interface used by the plugin.
+#
+# [*dispatch*]
+#  Dispatch the OVS DB interface link status value with configured plugin
+#  interval.
+#
+# [*ensure*]
+#  ensure param for collectd::plugin type.
+#  Defaults to 'ensure'
+#
+# [* interfaces *]
+#  List of interface names to be monitored by this plugin. If this option
+#  is not specified or is empty then all OVS connected interfaces
+#  on all bridges are monitored.
+#
+# [*manage_package*]
+#  If enabled, manages separate package for plugin
+#  Defaults to true
+#
+# [*send_notification*]
+#  If set to true, OVS link notifications (interface status and
+#  OVS DB connection terminate) are sent to collectd.
+#
+# [*package_name*]
+#  If manage_package is true, this gives the name of the package to manage.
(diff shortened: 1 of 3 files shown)


paramite:
Can this be merged now please, when all checks are passing?

paramite:
@alexjfisher This one is similar, can you please check it too? Thanks in advance.

templates/plugin/ovs_events.conf.erb
```diff
@@ -0,0 +1,20 @@
...
+<%- if @address %>
+  Address "<%= @address %>"
+<%- end -%>
+<%- if @socket %>
+  Socket "<%= @socket %>"
+<%- end -%>
+<%- if @interfaces %>
+  Interfaces "<%= @interfaces.join('" "') %>"
+<%- end -%>
+<%- if defined?(@send_notification) %>
```
alexjfisher: I'd have gone with `unless @send_notification.null?`, but I guess `if defined?` also works.

manifests/plugin/ovs_events.pp
```diff
@@ -0,0 +1,70 @@
...
+#
+# [*socket*]
+#  The UNIX domain socket path of OVS DB server JSON-RPC interface used
+#  by the plugin
+#
+class collectd::plugin::ovs_events (
+  Optional[String] $address            = undef,
+  Optional[Boolean] $dispatch          = undef,
+  String $ensure                       = 'present',
+  Optional[Array] $interfaces          = undef,
```
alexjfisher: Maybe better to use `Optional[Array[1]]`? or default to an empty array and adjust the template.

manifests/plugin/ovs_events.pp
```diff
@@ -0,0 +1,70 @@
...
+#
+# [*port*]
+#  TCP-port to connect to. Either a service name or a port number may be given.
+#
+# [*socket*]
+#  The UNIX domain socket path of OVS DB server JSON-RPC interface used
+#  by the plugin
+#
+class collectd::plugin::ovs_events (
+  Optional[String] $address            = undef,
```
alexjfisher: `Stdlib::Host` is available in recent stdlib.

alexjfisher:
@paramite LGTM.  I've left a couple of random comments, but these are very minor and subjective.

paramite:
@alexjfisher So as test suite discovered, "@send_notification.null?" does not work. I've added the other suggested changes.

manifests/plugin/ovs_events.pp
```diff
@@ -0,0 +1,70 @@
...
+#
+# [*port*]
+#  TCP-port to connect to. Either a service name or a port number may be given.
+#
+# [*socket*]
+#  The UNIX domain socket path of OVS DB server JSON-RPC interface used
+#  by the plugin
+#
+class collectd::plugin::ovs_events (
+  Optional[Stdlib::Host] $address      = undef,
```
bastelfreak: datatypes \o/

bastelfreak:
Thanks for the PR @paramite !

PR #746, opened 2018-01-18
feat: add package_keyserver parameter

marcdeop:
We have noticed sometimes puppet fails to provision our servers due to failing connectivity to pgp.mit.edu.

This MR allows you to choose the keyserver in case such an event happens.

Thanks to [Erasys GmbH](https://www.erasys.de/)

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -15,6 +15,7 @@ class collectd (
   $minimum_version                                                    = $collectd::params::minimum_version,
   $package_ensure                                                     = $collectd::params::package_ensure,
   $package_install_options                                            = $collectd::params::package_install_options,
+  String $package_keyserver                                           = $collectd::params::package_keyserver,
   $package_name                                                       = $collectd::params::package_name,
   $package_provider                                                   = $collectd::params::package_provider,
   $plugin_conf_dir                                                    = $collectd::params::plugin_conf_dir,
--- a/README.md
+++ b/README.md
@@ -77,6 +77,11 @@ On CentOS that will be EPEL:
 On Ubuntu that'll be the CollectD PPA:
 * https://launchpad.net/~collectd/+archive/ubuntu/collectd-5.5
 
+### Public key keyserver
+In case you need to change the server where to download the public key from for
+whatever reason (AKA: server is down) you can use the parameter
+`$package_keyserver`
+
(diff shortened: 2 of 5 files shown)


bastelfreak:
Thanks for the pr @marcdeop !

manifests/init.pp
```diff
@@ -15,6 +15,7 @@
   $minimum_version                                                    = $collectd::params::minimum_version,
   $package_ensure                                                     = $collectd::params::package_ensure,
   $package_install_options                                            = $collectd::params::package_install_options,
+  $package_keyserver                                                  = $collectd::params::package_keyserver,
```
bastelfreak: please add a datatype to the new param.

marcdeop: I tipbased and added the `String` datatype to the package_keyserver parameter.

I was considering using `Stdlib::Fqdn` but is not supported by the minimum stdlib library supported by this module (`4.13.1`)

bastelfreak:
@marcdeop can you bump the version of stdlib in one commit and use Stdlib::Fqdn in the existing commit? Bumping it is fine.

hadret:
@bastelfreak is there anything blocking this PR from being merged?

bastelfreak:
As written in the last comment:
Please switch from String to Stdlib::Fqdn. Bump the version boundary of stdlib in another commit (but within this PR).

marcdeop:
Thanks for the reminder about this PR @hadret 

@bastelfreak stdlib supports Stdlib::Fqdn as seen [here](https://github.com/puppetlabs/puppetlabs-stdlib#stdlibfqdn) but apparently the support is only on the master branch. The [relevant PR](https://github.com/puppetlabs/puppetlabs-stdlib/pull/842) was merged after the latest release (4.24.0) was published.

Should we just go ahead and merge this PR and at the same type create an [issue](https://github.com/voxpupuli/puppet-collectd/issues/new) to handle the stdlib::fqdn type as soon as its released by puppetlabs-stdlib?

bastelfreak:
Ah shit. I thought it's already released. I will merge this one as is. Can you spin up a new PR that uses the fqdn type,  but with a comment that it requires a new stdlib release?

marcdeop:
Done: #759

PR #759, opened 2018-03-06
feat: use Stdlib::Fqdn type for package_keyserver

marcdeop:
As discussed #746 here is a PR to fix the data type for package_keyserver.

This is cannot be merged until there is a release of puppetlabs-stdlib that include the proper datatype (it's only on master right now)

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -15,7 +15,7 @@ class collectd (
   $minimum_version                                                    = $collectd::params::minimum_version,
   $package_ensure                                                     = $collectd::params::package_ensure,
   $package_install_options                                            = $collectd::params::package_install_options,
-  String $package_keyserver                                           = $collectd::params::package_keyserver,
+  Stdlib::Fqdn $package_keyserver                                     = $collectd::params::package_keyserver,
   $package_name                                                       = $collectd::params::package_name,
   $package_provider                                                   = $collectd::params::package_provider,
   $plugin_conf_dir                                                    = $collectd::params::plugin_conf_dir,
--- a/metadata.json
+++ b/metadata.json
@@ -91,7 +91,7 @@
   "dependencies": [
     {
       "name": "puppetlabs/stdlib",
-      "version_requirement": ">= 4.13.1 < 5.0.0"
+      "version_requirement": ">= 4.25.0 < 5.0.0"
     },
     {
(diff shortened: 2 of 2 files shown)


bastelfreak:
I restarted the travis jobs. @marcdeop https://forge.puppet.com/puppetlabs/stdlib got released 3 days ago. Can you please bump the required version in metadata.json?

marcdeop:
Commit 7ae6079674cb9cc215aa91530ca440661ce8f104 bumps minimum stdlib version both in metadata.json and in .fixtures.yml so we can make sure that tests run the same way as the production code :-)

.fixtures.yml
```diff
@@ -4,6 +4,7 @@ fixtures:
       repo: 'git://github.com/puppetlabs/puppetlabs-apt'
     stdlib:
       repo: 'git://github.com/puppetlabs/puppetlabs-stdlib'
+      ref: "4.25.0"
```
bastelfreak: please don't bump it here. We always want to test against the master branch to catch unreleased errors early.

marcdeop: I removed it from the commit even though I don't understand the reason.
Which unreleased errors? From the stdlib library? That would not make much sense...

I can only see downsides on doing this: you are testing with code that it will be *different* from what the actual production code will be based on. How does that make sense?

bastelfreak: that's a complicated story. In the past our modules broke often because a breaking change was introduced into a dependency, back then we references releases in the .fixtures.yml. We pinned it to the lowest known working version. The puppet module tool always tries to download the latest version for dependencies.

tldr: We noticed that dependencies aren't compatible in the latest version only because it failed for our users and they reported it.

What we now do is a bit smarter. We always test against the master branches of our dependencies. So we notice during our regular PRs to our modules, if a dependency isn't working anymore. This is all related to our unit tests.

-----

Our acceptance tests are different. We start docker instances on travis, install dependencies from our metadata.json, execute puppet  code and check if everything   is working. We're simulating our users.

Combination of unit and acceptance tests are, in my opinion, the best way to address/track breaking changes.

This was a rather long explanation and I'm not a native speaker. If this text raises more questions than it answers please feel free to join our IRC channel #voxpupuli on freenode and we can talk about this.

PR #664, opened 2017-04-19
Add install_options for RedHat Family in sensors plugin

ziporah:
This patch adds the global specified installation options for the sensors plugin

PR #762, opened 2018-03-22
Install package with options in sensors and disk plugins

jfroche:
We need to be able to specify some options when installing plugins package (e.g. --enablerepo) when we are using Redhat.

Replacement for #664


bastelfreak:
Thanks for the PR @jfroche !

PR #767, opened 2018-03-30
Standardize file permissions, allow non-root

### LINKED ISSUE #719: Owner and modes for plugin files should reference ones set in collectd init
file and concat resource types have:
 ```  
 mode           => '0640',
 owner          => 'root',
```

These should be taken from new parameters in init.pp

juniorsysadmin:
This change allows .conf and .load file permission to be changed from a
single place, and potentially helps with non-root installations.

- Change root_group parameter to config_group
- Add config_owner and config_mode parameters
- collectd::plugin::filter::chain Change default permissions for the
  conf file from 0644 to 0640
- collectd::plugin::perl Replace missing root_user with config_owner
- collectd::plugin::perl Change default permissions for the
  configuration directory from 0755 to 0750
- collectd::plugin::perl::plugin Change default permissions for the
  conf file from 0644 to 0640
- collectd::plugin::python Change default permissions for the
  Python module directories from 0755 to 0750
- collectd::plugin::tail::file change default conf file permissions
  from 0644 to 0640
- collectd::typesdb Add config_mode and config_owner parameters

Fixes #719


juniorsysadmin:
Doc and test updates to be done later in another PR

bastelfreak:
Should we really do the doc update in another PR? Do we need to adjust tests?

juniorsysadmin:
> Should we really do the doc update in another PR? Do we need to adjust tests?

`root_group` is not documented at the moment, nor explicitly tested. I would prefer to do separate additions to all of our tests after further some major changes are done shortly, then a separate doc-update pass.

PR #769, opened 2018-03-30
Remove parameters from collectd::install class

juniorsysadmin:
Use variables from init.pp directly

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -37,9 +37,7 @@ class collectd (
 
   $collectd_version_real = pick_default($facts['collectd_version'], $minimum_version)
 
-  class { '::collectd::install':
-    package_install_options => $package_install_options,
-  }
+  class { '::collectd::install': }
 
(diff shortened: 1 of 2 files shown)


manifests/init.pp
```diff
@@ -37,9 +37,7 @@
 
   $collectd_version_real = pick_default($facts['collectd_version'], $minimum_version)
 
-  class { '::collectd::install':
-    package_install_options => $package_install_options,
-  }
+  class { '::collectd::install': }
```
bastelfreak: this should be converted to 'contain collectd::install'?

bastelfreak:
+1 from my side. Can you take a look at the inline comment? We should probably add this to our review guideline?

manifests/init.pp
juniorsysadmin: Yes it should, but I intend to do this in later PR as part of data type changes to init.pp

Issue #633, opened 2017-01-19
Reduce test runtimes


rnelson0:
It has been pointed out that tests on collectd take a long time to run, approx 15+ minutes on Travis CI and possibly longer. This makes adjusting a PR a test of patience. Average test runtimes for VP modules are in the 3-5 minute range making this 3-5x longer. We need to investigate a way to reduce the time.

We would like to solicit some solutions. Two that have already been discussed as low-hanging fruit are:

- [ ] Parallelized rspec tests
- [ ] Drop some OS support

We are investigating the first already. The second is a bit more controversial. Per the [current metadata.json](https://github.com/voxpupuli/puppet-collectd/blob/815bd53cbc963a1123a9e7bfbfca9c214d696ee6/metadata.json) most OSes are still supported. Only Debian 6 looks to be fully past end of life. However, the Enterprise Linux 5 family is coming up on End of Life (20170331) and some others may be nearing it as well. We loop over each supported OS in most of the tests, so removing even a single OS from the matrix should have some effect on the tests.

*We are proposing to drop Debian 6 support now.* We also plan to drop the EL5 family at the end of March.

That leads us to the more difficult target:

- [ ] Refactor tests for improved runtime without dropping coverage

We would like help from frequent contributors and users with this step. Specifically, highlighting some of the slowest tests, the easiest tests to refactor, and any unnecessary tests.

bastelfreak:
I'm fine with dropping ubuntu12.04, RHEL5 and Debian 6.

rnelson0:
Parallelization does seem to help, though the output is much different:
```
$ time be rake parallel_test

2296 examples, 1 failure

Took 533 seconds (8:53)
Tests Failed

real    8m56.077s
user    11m13.905s
sys     0m5.794s
```

```
$ time be rake test

Finished in 13 minutes 2 seconds (files took 16.3 seconds to load)
2296 examples, 1 failure

real    13m51.740s
user    13m32.719s
sys     0m6.207s
```

rnelson0:
#634 shows the reduced runtime. ~10 minutes vs ~16, hopefully less on average since one of the jobs seemed to hang longer than the others.. It gets 32 parallel threads on the Travis VMs, which makes it FAST. That's the positive.

[...]

PR #770, opened 2018-03-31
Use one operating system for most RSpec tests

juniorsysadmin:
The majority of Puppet classes and defined types in this module do not
have behaviour which changes depending on the operating system. Mock one
operating system for most tests, except for collectd_init_spec when
mocking multiple operating systems makes sense.

This attempts to address #633 , and should reduce test runtimes
considerably with little change in the confidence of tests actually
picking up problems.


bastelfreak:
I"m fine with this but we should wait with merging to get a bit feedback.

Issue #269, opened 2015-06-08
Under Debian, does not update `collectd-core` when require => 'latest'


rymndhng:
If you're trying to perform an upgrade on an existing collectd installation and set `require => 'latest'`, `collectd-core` does not get updated.

I would argue that this is more of an issue on the Debian side not specifying version requirements for `collectd-core` but it'd be great if this module would abstract this problem away.

txaj:
Could you give us a complete snippet to reproduce ? Aren't you looking to set latest here instead ?  https://github.com/pdxcat/puppet-module-collectd/blob/1830403421c03c10a00552075e164214e97c86da/manifests/init.pp#L16

rymndhng:
On initial apply of puppet manifest, it works fine.

It doesn't work properly when performing an upgrade on an existing `collectd` installation:
1. run puppet with  `collectd` resource with parameter `require => 'latest'`.
(This will download the latest version of `collectd` and `collectd-core`)
2. new version of `collectd` is released
3. re-run puppet
(updates `collectd` but *not* `collectd-core`)

txaj:
Well... 
```
# apt-cache show collectd|grep Depends
Depends: collectd-core, iptables, libc6 (>= 2.4), librrd4 (>= 1.3.0)
```

Is the mismatch between collectd and collectd-core causing issues ? If yes I'd also recommend to send this to Debian instead... Other than that I found unpractical to use ensure => latest anyway as it can cause surprise unplanned upgrades...

rymndhng:
Yeah, I understand where you're coming from. In our setup we had `ensure => 'installed'`, but on decision to use this module to update to the latest version, it's behavior did not match the expectation. 

What I mean is that why have `ensure => 'latest'` if it's functionality is broken (albeit I haven't tried other platforms, hopefully they don't have this issue as well). I've filed a [ticket](https://bugs.launchpad.net/ubuntu/+source/collectd/+bug/1463916) upstream.

What are your thoughts on patching this inside the `collectd` module while this gets resolved though, because without managing the `collectd-core` package currently on Debian, `ensure => 'latest'` is broken.

jyaworski:
I'd rather not have us code around OS packaging problems. However, we could ensure an array of packages on (I assume) <code>$::osfamily == 'Debian'</code>.

jyaworski:
Let me know if #406 solves this for you.

PR #773, opened 2018-04-01
Use data types, remove anchors

juniorsysadmin:
- Only allow Boolean for $purge
- Only allow Boolean for $recurse
- Only allow Integer for $write_queue_limit_low
- Only allow Integer for $write_queue_limit_high
- Don't allow undef for $manage_service
- Replace anchors with contain
- Remove some unecessary tests

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -1,60 +1,51 @@
 #
 class collectd (
-  $collectd_hostname                                                  = $collectd::params::collectd_hostname,
-  $conf_content                                                       = $collectd::params::conf_content,
-  $config_file                                                        = $collectd::params::config_file,
-  $fqdnlookup                                                         = $collectd::params::fqdnlookup,
-  $has_wordexp                                                        = $collectd::params::has_wordexp,
-  $include                                                            = $collectd::params::include,
-  $interval                                                           = $collectd::params::interval,
-  $internal_stats                                                     = $collectd::params::internal_stats,
-  $manage_package                                                     = $collectd::params::manage_package,
-  $manage_repo                                                        = $collectd::params::manage_repo,
-  Optional[Collectd::Manifests::Init] $ci_package_repo                = $collectd::params::ci_package_repo,
-  $manage_service                                                     = $collectd::params::manage_service,
-  $minimum_version                                                    = $collectd::params::minimum_version,
-  $package_ensure                                                     = $collectd::params::package_ensure,
-  $package_install_options                                            = $collectd::params::package_install_options,
-  Stdlib::Fqdn $package_keyserver                                     = $collectd::params::package_keyserver,
-  $package_name                                                       = $collectd::params::package_name,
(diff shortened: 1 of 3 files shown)


manifests/init.pp
```diff
@@ -1,60 +1,51 @@
...
+  Boolean $internal_stats                              = $collectd::params::internal_stats,
+  Boolean $manage_package                              = $collectd::params::manage_package,
+  Boolean $manage_repo                                 = $collectd::params::manage_repo,
+  Optional[Collectd::Manifests::Init] $ci_package_repo = $collectd::params::ci_package_repo,
+  Boolean $manage_service                              = $collectd::params::manage_service,
+  String $minimum_version                              = $collectd::params::minimum_version,
+  String $package_ensure                               = $collectd::params::package_ensure,
+  Optional[Array] $package_install_options             = $collectd::params::package_install_options,
+  Stdlib::Fqdn $package_keyserver                      = $collectd::params::package_keyserver,
+  Variant[String[1],Array[String[1]]] $package_name    = $collectd::params::package_name,
```
juniorsysadmin: #269 Appears to still affect recent Debian-variants

spec/classes/collectd_init_spec.rb
```diff
@@ -9,20 +9,13 @@
 
       options = os_specific_options(facts)
       context 'with all defaults' do
-        it { is_expected.to contain_class('collectd') }
```
bastelfreak: why did you remove those two?

spec/classes/collectd_init_spec.rb
```diff
@@ -9,20 +9,13 @@
...
         it { is_expected.to compile.with_all_deps }
-        it { is_expected.to contain_anchor('collectd::begin') }
-        it { is_expected.to contain_anchor('collectd::end') }
         it { is_expected.to contain_file('collectd.conf').without_content }
         it { is_expected.to contain_file('collectd.d').with_ensure('directory') }
         it { is_expected.to contain_file_line('include_conf_d').with_ensure('absent') }
         it { is_expected.to contain_file_line('include_conf_d_dot_conf').with_ensure('present') }
         it { is_expected.to contain_package(options[:package]).with_ensure('present') }
         it { is_expected.to contain_package(options[:package]).with_install_options(nil) }
-        it { is_expected.to contain_class('collectd::install').that_comes_before('Class[collectd::config]') }
```
bastelfreak: those checks should still be valid, since we use ->?

spec/classes/collectd_init_spec.rb
juniorsysadmin: When inside `init.pp` we know that a class called collectd will always be there. I just didn't see any value to the test.
Same with params.

spec/classes/collectd_init_spec.rb
juniorsysadmin: Valid, but again I don't think it is useful, as we don't have any conditional logic which will result in the order of classes changing.

PR #775, opened 2018-04-03
collectd::plugin::network - Add data types

juniorsysadmin:
- $port is now required to be an Integer type, not String

### DIFF
--- a/types/network/securitylevel.pp
+++ b/types/network/securitylevel.pp
@@ -1,0 +1,2 @@
+#
+type Collectd::Network::SecurityLevel = Enum['Encrypt', 'Sign', 'None']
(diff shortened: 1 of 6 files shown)


types/network/securitylevel.pp
```diff
@@ -0,0 +1,2 @@
+#
+type Collectd::Network::SecurityLevel = Enum['Encrypt', 'Sign', 'None']
```
bastelfreak: should we namespace this or use a shorter name, like: Collectd::SecurityLevel? That would keep the lines in the Puppet files shorter.

juniorsysadmin: I was trying to keep it consistent...I'm not sure that you can avoid many long lines when using types though.

PR #792, opened 2018-05-18
Remove array sorting in python module template (#586)

### LINKED ISSUE #586: Array sorted in python module
Hello,

In the python module, there is a template used to write config files for collectd.

https://github.com/voxpupuli/puppet-collectd/blob/master/templates/plugin/python/module.conf.erb#L7

Values are written in an order (with a sort). I understand : this is puppet's logic.
But, I think it shouldn't be like this for arrays (see line 7). We're loosing the array's order to an alphabetic order :/
This is a problem because the array in puppet's config is not the same in the file (order has been changed).
For my part, I removed the sort and replaced this code :

  <%- if value.is_a?(Array) -%>
  <%- value.sort.each do |v| -%>
    ...
  <%- end -%>

By this :

  <%- if value.is_a?(Array) -%>
  <%- value.each do |v| -%>           # no sort
    ...
  <%- end -%>

in this file : https://github.com/voxpupuli/puppet-collectd/blob/master/templates/plugin/python/module.conf.erb#L7

Thanks for your attention,

Gilles

jlutran:

    This PR removes array sorting in python module template
    We need to keep the array order provided by the config parameter of the collectd::plugin::python::module function parameter instead of an alphabetic order.
    Please see #586 for details.

    Fixes #586

### DIFF
--- a/templates/plugin/python/module.conf.erb
+++ b/templates/plugin/python/module.conf.erb
@@ -6,7 +6,7 @@
   <Module "<%= @module %>">
   <%- configuration.sort.each do |key,value| -%>
   <%- if value.is_a?(Array) -%>
-  <%- value.sort.each do |v| -%>
+  <%- value.each do |v| -%>
     <%= key %> <% if !!v == v %><%= v %><% else %>"<%= Shellwords.split(v).join('" "') %>"<% end %>
   <%- end -%>
   <%- elsif value.is_a?(Hash) -%>


bastelfreak:
Fixes https://github.com/voxpupuli/puppet-collectd/issues/586

bastelfreak:
Thanks @jlutran !

PR #798, opened 2018-05-25
apache: sync options and add tests

sileht:
apache: sync options and add tests

This change adds test coverage for the apache plugins.

And two missing options SSLCiphers and Timeout.


spec/classes/collectd_plugin_apache_instance_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    SSLCiphers "TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_GCM_SHA256"
+    Timeout 120
+  </Instance>
+</Plugin>
+EOS
+          is_expected.to contain_file('apache-instance-site2.conf').with(
+            content: content_instance_file,
+            path: "#{options[:plugin_conf_dir]}/25-apache-instance-site2.conf"
+          )
+        end
```
bastelfreak: could you add another test to verify that it compiles with all dependencies?

spec/classes/collectd_plugin_apache_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+        context ':manage_package => false on osfamily => RedHat' do
+          let :params do
+            {
+              manage_package: false
+            }
+          end
+
+          it 'Will not manage collectd-apache' do
+            is_expected.not_to contain_package('collectd-apache').with(ensure: 'present',
+                                                                       name: 'collectd-apache')
```
bastelfreak: same as above

bastelfreak:
Hi @sileht, thanks for the PR. Can you take a look at the inline comments and the failing travis job?

spec/classes/collectd_plugin_apache_instance_spec.rb
sileht: I have added check that required class are included, is that enough ?

bastelfreak: please add a `is_expected.to compile.with_all_deps`

sileht: gotcha, done

spec/classes/collectd_plugin_apache_spec.rb
sileht: done

PR #800, opened 2018-05-25
plugin: add battery plugin

sileht:
this change adds the battery plugin


manifests/plugin/battery.pp
```diff
@@ -0,0 +1,17 @@
+# https://collectd.org/wiki/index.php/Plugin:Uptime
+class collectd::plugin::battery (
+  $ensure   = 'present',
```
bastelfreak: please add a datatype for `$ensure` as well.

spec/classes/collectd_plugin_battery_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+# Generated by Puppet
+<LoadPlugin battery>
+  Globals false
+</LoadPlugin>
+
+<Plugin "battery">
+</Plugin>
+
+EOS
+
```
bastelfreak: please also add a `is_expected.to compile.with_all_deps` test

templates/plugin/battery.conf.erb
```diff
@@ -0,0 +1,11 @@
+<Plugin "battery">
```
bastelfreak: for new templates we agreed on epp only: https://voxpupuli.org/docs/#reviewing-a-module-pr
Can you please convert this to epp?

manifests/plugin/battery.pp
```diff
@@ -0,0 +1,17 @@
...
+  $interval = undef,
+  Optional[Boolean] $values_percentage = undef,
```
bastelfreak: can we instead default to true or false and don't mark it as optional?

sileht: done

templates/plugin/battery.conf.erb
sileht: done

manifests/plugin/battery.pp
sileht: done

spec/classes/collectd_plugin_battery_spec.rb
sileht: done

PR #801, opened 2018-05-25
plugin: update cpu for collectd 5.8

sileht:
This change adds cpu plugin options added in collectd 5.8


manifests/plugin/cpu.pp
```diff
@@ -5,6 +5,8 @@
   Boolean $reportbycpu              = true,
   Boolean $valuespercentage         = false,
   Boolean $reportnumcpu             = false,
+  Boolean $reportgueststate         = false,
```
bastelfreak: datatypes \o/

templates/plugin/cpu.conf.erb
```diff
@@ -6,5 +6,9 @@
 <% if scope.lookupvar('collectd::collectd_version_real') and scope.function_versioncmp([scope.lookupvar('collectd::collectd_version_real'), '5.6']) >= 0 -%>
   ReportNumCpu <%= @reportnumcpu %>
 <% end -%>
+<% if scope.lookupvar('collectd::collectd_version_real') and scope.function_versioncmp([scope.lookupvar('collectd::collectd_version_real'), '5.8']) >= 0 -%>
```
bastelfreak: `if @collectd::collectd_version_real` should work as well. scope.lookupvar() is legacy stuff to be compatible with very old puppet versions. It isn't needed anymore

templates/plugin/cpu.conf.erb
```diff
@@ -6,5 +6,9 @@
...
+    ReportGuestState <%= @reportgueststate %>
+    SubtractGuestState <%= @subtractgueststate %>
+<% end -%>
```
bastelfreak: are you interested in converting this erb template into epp? This is not required, but we would be very happy about it :)

bastelfreak:
Thanks for the PR @sileht ! I added some inline comments.

templates/plugin/cpu.conf.erb
sileht: I don't get why this is even needed, I have checked how this variable is built, and if the fact return null or undef, we pick the minimal_version. So it can never be null or I miss something ?

templates/plugin/cpu.conf.erb
sileht: It's OK, the file small enought, I done it :)

templates/plugin/cpu.conf.erb
bastelfreak: I'm not 100% sure if that's still needed. That's something we could check and cleanup.

PR #802, opened 2018-05-28
Add amqp1 plugin support

paramite:
This patch adds possibility to configure amqp1 plugin added recently
to collectd [1].

[1] https://github.com/collectd/collectd/pull/2618

### DIFF
--- a/spec/classes/collectd_plugin_amqp1_spec.rb
+++ b/spec/classes/collectd_plugin_amqp1_spec.rb
@@ -1,0 +1,98 @@
+require 'spec_helper'
+
+describe 'collectd::plugin::amqp1', type: :class do
+  on_supported_os(baseline_os_hash).each do |os, facts|
+    context "on #{os} " do
+      let :facts do
+        facts
+      end
+
+      let :pre_condition do
+        'include collectd'
+      end
+
+      options = os_specific_options(facts)
+      context ':ensure => present' do
+        let :params do
+          { ensure: 'present' }
+        end
+
+        it { is_expected.to compile.with_all_deps }
+        it { is_expected.to contain_collectd__plugin('amqp1') }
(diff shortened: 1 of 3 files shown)


spec/classes/collectd_plugin_amqp1_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+                  graphite_escape_char: '_',
+                  graphite_separate_instances: false,
+                  graphite_always_append_ds: false,
+                  graphite_preserve_separator: false
+                }
+            ]
+          }
+        end
+
+        it { is_expected.to contain_file('amqp1.load').with(content: %r{<Transport "transport">}) }
```
bastelfreak: can you add a `it { is_expected.to compile.with_all_deps }`

templates/plugin/amqp1.conf.erb
```diff
@@ -0,0 +1,46 @@
+<Plugin amqp1>
```
bastelfreak: can you please provide this as an epp and not erb template? We agreed on epp some time ago and documented it in our review guideline at https://voxpupuli.org/docs/#reviewing-a-module-pr

spec/classes/collectd_plugin_amqp1_spec.rb
paramite: Sure will do.

templates/plugin/amqp1.conf.erb
paramite: Ok, didn't notice this as all templates are in ERB. Will rewrite the template then.

bastelfreak:
@paramite can you take a look at the failing travis jobs?

paramite:
Yes, I'mrreally trying, but EPP and syntax check is giving me hard times :).

paramite:
@bastelfreak Done.

PR #804, opened 2018-05-30
Modularize python plugin configuration

traylenator:

Previously a `python-config.conf`

```ApacheConf
<Plugin "python">
  ModulePath "/usr/lib/python2.7/site-packages"
  ModulePath "/usr/libexec/sensors"
  LogTraces true
  Interactive false
  Import "moduleA"

  <Module "moduleA">
    Value "60"
    Value "100"
  </Module>

  Import "moduleB"

  <Module "moduleB">
    Value "120"
  </Module>

</Plugin>
```

This patch rewrites the configuration to

```ApacheConf
<Plugin "python">
  ModulePath "/usr/lib/python2.7/site-packages"
  ModulePath "/usr/libexec/sensors"
  LogTraces true
  Interactive false
</Plugin>
Include /etc/collectd.d/python.d/*.conf
```
And creates two files in the directory, one per python module, e.g. `/etc/collect.d/python.d/moduleA.conf` with

```ApacheConf
<Module Python>
  Import "moduleA"
  <Module "moduleA">
    Value "60"
    Value "100"
  </Module>
</Module Python>
```

As far as I can ascertain these multiple `Module Python` blocks are all loaded together so no
actual difference.

In addition these per python module files are maintained as a concat so it easy to inject configuration
arbitary values into the Module Configration with a defined type.

My use case was this. Create a defined type to configure indidual bits of  array of parameters to a defined type.
One instance defining `60` the other `100`

### DIFF
--- a/templates/plugin/python/module_header.conf.erb
+++ b/templates/plugin/python/module_header.conf.erb
@@ -1,3 +1,4 @@
+<Plugin "python">
   <%- require 'shellwords' -%>
   Import "<%= @_module_import %>"
 
@@ -17,5 +18,4 @@
     <%= key %> <% if !!value == value %><%= value %><% else %>"<%= Shellwords.split(value).join('" "') %>"<% end %>
   <%- end -%>
   <%- end -%>
-  </Module>
   <%- end -%>
--- a/README.md
+++ b/README.md
@@ -1428,6 +1428,8 @@ non-default value.
   collectd with the name of the exception and the message (default: `false`)
 * `conf_name` name of the file that will contain the python module configuration
   (default: `python-config.conf`)
+* `python_conf_dir` name of directory inside which individual python plugins will write
+  their configuration to. (default: `python.d`)
 
  See [collectd-python documentation](https://collectd.org/documentation/manpages/collectd-python.5.shtml)
  for more details.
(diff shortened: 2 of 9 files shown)


traylenator:
I might try writing a python plugin acceptance test which might be good before this.

traylenator:
#805 should go in first.

templates/plugin/python/module_header.conf.erb
```diff
@@ -1,3 +1,4 @@
+<Plugin "python">
```
bastelfreak: maybe it's now time to convert this to an epp template?

traylenator: I might tackle but best as a separate MR I think.

traylenator:
While trying to fix another bug I realised this will go completely bad for the case where multiple instances of python plugin are are configured as an array of configuration.

```puppet
class { 'collectd::plugin::python':
  modulepaths => ['/usr/share/collectd/python'],
  modules     => {
    'another-module' => {
      'config'        => [{'Verbose' => 'true'}, {'Verbose' => 'false'}],
    }
  }
  logtraces   => true,
  interactive => false,
}
```
I'll add an acceptance test for that case first.
But the end goal is now harder to achieve.

bastelfreak:
@traylenator can you please rebase?

traylenator:
This is obsoleted by #821 which is simpler and actually works.

PR #805, opened 2018-05-30
A collectd::plugin::python acceptance test

traylenator:
A collectd::plugin::python acceptance test

### DIFF
--- a/README.md
+++ b/README.md
@@ -1407,6 +1407,9 @@ class { 'collectd::plugin::protocols':
 ```
 
 #### Class: `collectd::plugin::python`
+The plugin uses a fact `python_dir` to find the python load path for modules.
+python must be installed as a pre-requisite for the this fact to give a
+non-default value.
 
 * `modulepaths` is an array of paths where will be Collectd looking for Python
   modules, Puppet will ensure that each of specified directories exists and it
--- a/spec/acceptance/class_plugin_python_spec.rb
+++ b/spec/acceptance/class_plugin_python_spec.rb
@@ -1,0 +1,71 @@
+require 'spec_helper_acceptance'
+
+describe 'collectd::plugin::python class' do
+  context 'basic parameters' do
+    # Using puppet_apply as a helper
+    it 'works idempotently with no errors' do
+      pp = <<-EOS
+      class{'collectd::plugin::python':
+      }
+      EOS
+      # Run 3 times since the collectd_version
+      # fact is impossible until collectd is
+      # installed. This acceptance test happens
+      # to be first.
+      apply_manifest(pp, catch_failures: true)
+      apply_manifest(pp, catch_failures: true)
+      apply_manifest(pp, catch_changes: true)
+    end
+
+    describe service('collectd') do
+      it { is_expected.to be_running }
+    end
+  end
+
+  context 'trivial module' do
+    it 'works idempotently with no errors' do
+      pp = <<-EOS
+      if $facts['os']['family'] == 'Debian' {
+        # for collectdctl command
+        package{'collectd-utils':
+	  ensure => present,
+        }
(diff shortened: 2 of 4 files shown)


traylenator:
ubuntu 16 fails because collectd does not pull in python , only python-lib.

So the fact `python_dir` defaults to 

`/usr/local/lib/python2.7/dist-packages`

but this can't be made because 

`/usr/local/lib/python2.7`

does not exist until python-lib. 

In the real world this never happens probably because something pulls in python.

bastelfreak:
so python is a soft dependency? Can you document it in the README.md and install it during the acceptance tests?

traylenator:
It's really a hard dependency so added the python install the collectd::python::module.

manifests/plugin/python.pp
```diff
@@ -47,6 +49,8 @@
     globals  => $globals,
   }
 
+  ensure_packages([$python_package], { ensure => $python_package_ensure })
```
bastelfreak: we should probably wrap this into a $manage_python condition. There are distributions where 'python' defaults to python 3, others default to version 2. Do we need to catch that here as well?

traylenator: Collectd is always built against exactly one python. It is the `python` package untill we have a counter example.

Happy to add a manage_python boolean. It will help people who install python in another funky way.

manifests/plugin/python.pp
```diff
@@ -1,19 +1,22 @@
...
+  Boolean $globals              = true,
+  Boolean $interactive          = false,
+  $interval                     = undef,
+  Boolean $logtraces            = false,
+  $manage_package               = undef,
+  Array $modulepaths            = [],
+  Hash $modules                 = {},
+  $order                        = '10',
+  $conf_name                    = 'python-config.conf',
+  Boolean $manage_python        = true,
```
bastelfreak: I'm currently unsure if this should default to true. This probably breaks the catalog for many people. Should it default to false and be documented in the README.md?

vinzent: +1 to not install/manage prerequisites by default.

traylenator:
```
collectd::plugin::python
```

cannot work if python is not installed. What's the difference between a pre-requisite and a requirement?
Package is added with 'ensure_packages' to reduce impact to people catalogs.

Happy to change of course.

traylenator: I'll change because it's easier to do and get through. It's only ubunutu and I never use that so...

Think I already answered the python version comment above.

> ensure_packages works only if this class gets evaluated at the end.

The other class probably needs fixing if it does not ensure_resource on such a fundamental package.
Essentially this is saying it's never okay to install python but given ubunut 16 docker images don't have python this is a `new` necessity.

Of course there's a total alternate method to this. Remove the facter fact that depends on a python interpreter to work and just set sensible values for *python_sitelib* on different OSes.

manifests/plugin/python.pp
```diff
@@ -47,6 +50,13 @@
     globals  => $globals,
   }
 
+  if $manage_python {
+    ensure_packages([$python_package], { ensure => $python_package_ensure })
```
juniorsysadmin: I don't like this because it's really a soft dependency and is not some kind of separate `collectd_python` package. For example, if a plugin required mongodb, apache or ruby would it be appropriate to manage installation in this module? That being said, `ensure_packages` is the best way to do it.

traylenator: It's slightly different because it's the fact and puppet manifest that requires 'python'. collectd itself does not require it.
Anyway especially with the default false the code is basically pointless so lets just chop it.

traylenator:
Changed after approval, still okay?

PR #807, opened 2018-05-31
add centos 6 acceptance tests

bastelfreak:


bastelfreak:
The tests fail because the epel repository has a way too old collectd version (4.10).

PR #808, opened 2018-05-31
drop centos 6 support

bastelfreak:
CentOS 6 is too old, the provided packages in epel don't work anymore
with the module. Some failures can be seen at https://github.com/voxpupuli/puppet-collectd/pull/807

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -3,28 +3,24 @@
     {
       "operatingsystem": "RedHat",
       "operatingsystemrelease": [
-        "6",
         "7"
       ]
     },
     {
       "operatingsystem": "CentOS",
       "operatingsystemrelease": [
-        "6",
         "7"
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


dannygoulder:
oh noes.  this makes me very sad to see this merged! :cry: I've been building collectd from the instructions [here](https://github.com/collectd/collectd/blob/master/contrib/redhat/collectd.spec) and successfully used the module with 5.x versions on RHEL6 for a year or two now.  You could also get them from https://pkg.ci.collectd.org/rpm/collectd-5.8/epel-6-x86_64/.

I have a relatively large fleet of RHEL6 boxes managed by this module - would it be OK to submit PRs in case any functionality there gets broken?

bastelfreak:
@dannygoulder sure! As long as it works, has tests and the OS gets support upstream we are happy to review and merge.

PR #810, opened 2018-05-31
mysql: Add some missing options

sileht:
mysql: Add some missing options

This change adds some missing mysql options:
* Alias/ConnectionTimeout have been introduced in 5.5
* SSL* have been introduced in 5.6

I didn't convert the template to epp format, since epp in "define"
need a lot of work (all template variables must be parametized).


bastelfreak:
Thanks @sileht !

PR #811, opened 2018-06-05
add hugepages plugin

sileht:
Add hugepages plugin

In collectd 5.7, the hugepages plugin have been added.

This adds it in puppet-collectd.


bastelfreak:
Thanks @sileht !

PR #812, opened 2018-06-05
Add intel_rdt plugin

sileht:
Add intel_rdt plugin

intel_rdt plugin have been added collectd 5.7

This change adds it to puppet-collectd.


manifests/plugin/intel_rdt.pp
```diff
@@ -0,0 +1,42 @@
...
+#  represents a list of cores in a group. Allowed formats are: 0,1,2,3
+#  0-10,20-18 1,3,5-8,10,0x10-12
+#  If an empty string is provided as value for this field default cores
+#  configuration is applied - a separate group is created for each core.
+#  Defaults to ""
+#
+class collectd::plugin::intel_rdt (
+  Enum['present', 'absent'] $ensure   = 'present',
+  Optional[Integer] $interval         = undef,
+  Array[String] $cores                = [],
```
bastelfreak: This allows empty string as well. Can you update it to `Array[String[1]]`? That enforces a String length of at least one.

sileht: done

PR #813, opened 2018-06-05
Add numa plugin

sileht:
Add numa plugin

Numa plugin exists since 5.1, this change add it to puppet-collectd.


bastelfreak:
Thanks for the PR @sileht !

PR #814, opened 2018-06-05
swap: add ReportIO option

sileht:
swap: add ReportIO option

Add the new ReportIO option of collectd 5.8.


manifests/plugin/swap.pp
```diff
@@ -6,13 +6,14 @@
   Boolean $reportbytes      = true,
   Boolean $valuesabsolute   = true,
   Boolean $valuespercentage = false,
+  Boolean $reportio         = true,
```
juniorsysadmin: For new options, I personally would prefer that they be `Optional[Boolean]` so that backward compatibility is preserved and we can eventually drop all of the `if collected::version` conditional logic.

sileht: Sorry I'm lost, I have done that in the past and I have been asked to do the reverse https://github.com/voxpupuli/puppet-collectd/pull/800#discussion_r191935496

Can you see with @bastelfreak what you really want before I go further.

bastelfreak: From [our review guide](https://voxpupuli.org/docs/#reviewing-a-module-pr): 
> Are there new params with datatype Boolean? The default value is a tricky decision which needs careful reviewing. Sometimes a True/False is the better approach, sometimes undef

From an interface perspective it isn't a breaking change because the new parameter has a default value. Also it isn't written directly to the config file, only if collectd has the correct version, so the behavior isn't a breaking change either.

----

Instead of the collectd version check we could ensure it isn't undef. I'm not sure if we gain much when the parameter is undef. I'm a bit unsure here and would like to get more feedback from others.

PR #815, opened 2018-06-05
turbotstat: Add LogicalCoreNames option

sileht:
turbotstat: Add LogicalCoreNames option

Since collectd-5.7, we can configure the name convention for core names.

This change adds this.


templates/plugin/turbostat.conf.epp
```diff
@@ -0,0 +1,25 @@
+<Plugin turbostat>
+<% if $::collectd::plugin::turbostat::core_c_states { -%>
```
juniorsysadmin: Drop the `$::` - it's not needed

sileht: I have dropped the useless '::', but the '$' is required.

manifests/plugin/turbostat.pp
```diff
@@ -7,11 +7,12 @@
   Optional[Boolean]         $digital_temperature_sensor  = undef,
   Optional[Integer]         $tcc_activation_temp         = undef,
   Optional[Integer]         $running_average_power_limit = undef,
+  Optional[Boolean]         $logical_core_names          = undef,
```
bastelfreak: datatypes \o/

PR #816, opened 2018-06-12
ping: set parameters type and add Size

sileht:
ping: set parameters type and add Size

This change:
* adds the parameter type for all options
* adds the missing option size
* converts template to epp


manifests/plugin/ping.pp
```diff
@@ -1,14 +1,15 @@
...
   Array $hosts,
-  $ensure         = 'present',
-  $manage_package = undef,
-  $interval       = undef,
-  $timeout        = undef,
-  $ttl            = undef,
-  $source_address = undef,
-  $device         = undef,
-  $max_missed     = undef,
+  Enum['present', 'absent'] $ensure = 'present',
```
bastelfreak: datatypes \o/

manifests/plugin/ping.pp
```diff
@@ -26,6 +27,6 @@
   collectd::plugin { 'ping':
     ensure   => $ensure,
     interval => $interval,
-    content  => template('collectd/plugin/ping.conf.erb'),
+    content  => epp('collectd/plugin/ping.conf.epp'),
```
bastelfreak: epp \o/

PR #818, opened 2018-06-13
Fixes #662 accept numbers into configuration file

### LINKED ISSUE #662: collectd::plugin::python::module does not support number options
This module (via module.conf.erb) only supports boolean and string options.

When a number is given (e.g. config => [{ 'Port' => 1234}]), Puppet fails with "undefined method scan for 1234:Fixnum".

When a string is given (e.g. config => [{ 'Port' => '1234'}]), we end up with a string in python-config.conf.

Please add support for numbers: they should be rendered as they are, with no double quotes around.

traylenator:

With a configuration of

```puppet
collectd::python::python{'foo':
  modules => { 'foo' => { config => [{'value' => 4}]
                        }
             }
}
```

Puppet fails with `undefined method scan for 4:Fixnum`.

Integers and floats are now accepted and displayed as numbers.

Fixes #662

### DIFF
--- a/templates/plugin/python/module.conf.erb
+++ b/templates/plugin/python/module.conf.erb
@@ -7,14 +7,16 @@
   <%- configuration.sort.each do |key,value| -%>
   <%- if value.is_a?(Array) -%>
   <%- value.each do |v| -%>
-    <%= key %> <% if !!v == v %><%= v %><% else %>"<%= Shellwords.split(v).join('" "') %>"<% end %>
+    <%= key %> <% if !!v == v %><%= v %><% elsif v.is_a?(Numeric) %><%= v.to_s %><% else %>"<%= Shellwords.split(v).join('" "') %>"<% end %>
   <%- end -%>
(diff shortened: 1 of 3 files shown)


templates/plugin/python/module.conf.erb
```diff
@@ -7,14 +7,16 @@
   <%- configuration.sort.each do |key,value| -%>
   <%- if value.is_a?(Array) -%>
   <%- value.each do |v| -%>
-    <%= key %> <% if !!v == v %><%= v %><% else %>"<%= Shellwords.split(v).join('" "') %>"<% end %>
+    <%= key %> <% if !!v == v %><%= v %><% elsif v.is_a?(Numeric) %><%= v.to_s %><% else %>"<%= Shellwords.split(v).join('" "') %>"<% end %>
```
bastelfreak: it probably makes sense to convert this to epp

traylenator: See #819

PR #821, opened 2018-06-20
Cleanup how python modules are configured

### LINKED ISSUE #819: Hash, Array entries in config to python plugin ?

- Puppet: all
- Ruby: all
- Distribution: all
- Module version: 9.0.2-rc0

```
collectd::plugin::python{'extreme':
  modules => { 'extreme' => { 'config' => [{'Verbose' => true,
                                                                  'Array' => ['a', 'b', 'c'],
                                                                  'Hash' => {'d' => 'e', 'f' => 'g' },
                                                                }],
                                              }
                       }
}
```                     

## What are you seeing
`python-config.conf` contains

```config
Import "extreme"

<Module "extreme">
    Array "a"
    Array "b"
    Array "c"
    Hash d "e"
    Hash f "g"
    Verbose true
</Module>
```

The action is not documented so it is not obvious especially what the `Hash` should do.

If I had to guess.

```config
<Module "extreme">
    Array "a" "b" "c"
    <Hash>
      d "e"
      f  "g"
    </Hash>
    Verbose true
</Module>
```

## Any additional information you'd like to impart
Certainly the `hash` value I don't understand, I guess no one uses it or I am missing
the use case.
For the `array` value collectd modules are normally happy with either option but I would
suggest the 2nd is more natural.

Checking before contemplating `.epp`.

traylenator:
Convert templates for `collectd::module::python` to `epp`

Example

```puppet
collectd::plugin::python::module{'example':
  config => [{'Verbose' => true,
              'Values'  => ['abc',def'],
              'Name     => 'My Name',
              'Limit    => 3.4
             },
            ],
  }
}
```

This will now render as

```apache
Import "example"
<Module "example">
  Verbose true
  Values "abc" "def"
  Name "My Name"
  Limit 3.4
</Module>
```

Specifying a hash value in the `config` paramater will now result in a type
failure. Previously it would add nonsense to the configuration file.
e.g

```puppet
collectd::plugin::python::module{'example':
  config => [{'Values' => {'abc' => 'def'}}],
}
```
will now give a type error.

To specify a list of quoted strings use an array. Previously
the following could be used.

```puppet
collectd::plugin::python::module{'example':
  config => [{'Values' => '"abc" "def"',
              'Name'   => '"My Name"',
            }],
}
```

to acheive the same result use

```puppet
collectd::plugin::python::module{'example':
  config => [{'Values' => ['abc','def'],
              'Name'   => 'My Name',
            }],
}
```

to render

```apacheconf
Import "example"
<Module "example">
  Values "abc" "def"
  Name "My Name"
</Module>
```

It is now possible to specify multiple instances of a single python
module by using `collectd::plugin::python::module` twice. e.g.

```puppet
collectd::plugin::python::module{'exampleA':
  module => 'example',
  config => [{'Values' => 'abc'],}
}
collectd::plugin::python::module{'exampleB':
  module => 'example',
  config => [{'Values' => 'def'],}
}
```

Fixes #819


spec/classes/collectd_plugin_python_spec.rb
```diff
@@ -74,7 +74,7 @@
           end
         end
 
-        context ':ensure => present and configure elasticsearch module' do
+        context ':ensure => present and configure Elasticsearch module' do
```
traylenator: change should not be there.

manifests/plugin/python.pp
```diff
@@ -79,7 +79,7 @@
 
   concat::fragment { 'collectd_plugin_python_conf_header':
     order   => '00',
-    content => template('collectd/plugin/python/header.conf.erb'),
+    content => epp('collectd/plugin/python/header.conf'),
```
juniorsysadmin: Please make the extension .epp

traylenator: It's weird how it even works without. .epp added to that and other two places now.

bastelfreak:
thanks @traylenator !

PR #823, opened 2018-06-28
Add ipc plugin

sileht:
Add ipc plugin

Collectd plugin ipc have been added in Collectd 5.5.

This change adds its support to puppet-collectd


sileht:
@bastelfreak can you take a look ?

bastelfreak:
Thanks @sileht !

PR #826, opened 2018-07-09
Add table plugin

sileht:


manifests/plugin/table.pp
```diff
@@ -0,0 +1,24 @@
+# https://collectd.org/wiki/index.php/Chains
+class collectd::plugin::table (
+  Enum['present', 'absent'] $ensure    = 'present',
+  Integer $order                       = 10,
+  Hash[String, Struct[{
```
bastelfreak: Thanks for this awesome and detailed datatype! Can you please abstract it into its own datatype in https://github.com/voxpupuli/puppet-collectd/tree/master/types ?

manifests/plugin/table.pp
```diff
@@ -0,0 +1,24 @@
...
+    'instance'  => Optional[String],
+    'results'   => Array[Struct[{
+      'type'            => String,
+      'values_from'     => Array[Integer, 1],
+      'instance_prefix' => Optional[String],
+      'instances_from'  => Optional[Array[Integer, 1]]
+    }], 1]
+  }], 1] $tables,
+) {
+  include ::collectd
```
bastelfreak: please don't add the leading :: infront of a class name. This isn't needed anymore since we don't support puppet 3.

templates/plugin/table.conf.epp
```diff
@@ -0,0 +1,27 @@
+<Plugin "table">
```
bastelfreak: yey epp template \o/

templates/plugin/table.conf.epp
```diff
@@ -0,0 +1,27 @@
...
+<% $collectd::plugin::table::tables.each |$path, $table| { -%>
+  <Table "<%= $path %>">
+<% unless $table['plugin'] =~ Undef { -%>
```
bastelfreak: `if $table['plugin']` is the same as `unless $table['plugin'] =~ Undef`. I think the if notation is shorter and better to read?

bastelfreak:
Thanks for your work @sileht! can you please take a look at the inline comments I made?

templates/plugin/table.conf.epp
sileht: if plugin is set `""`, the `if` version will evaluation to `false` while the unless to `true`.

manifests/plugin/table.pp
sileht: Done, old habbit.

manifests/plugin/table.pp
sileht: Done, awesome I was not aware we can declare type there :)

PR #828, opened 2018-07-10
release 10.0.0

bastelfreak:

PR #830, opened 2018-07-11
write_riemann plugin fixes and enhancements

smortex:
The `write_riemann` plugin has some problems that are fixed by this problem report:

1. It is not possible to configure collectd to send events to multiple Riemann servers (now accept an array of nodes);
1. Some parameter names have an inconsistent extra "riemann_" at the beginning of their names (`riemann_host` and `riemann_port` where renamed to `host` and `port`);
1. Not many parameters types are checked (all of them are now tested);
1. The module plugin has some default values which are different from upstream (`protocols` and `store_rates` had wrong default value. Do not set default values to the plugin parameters, and only output those that where explicitly set by the end-user.  This allows collectd to use the default value if a parameter was not set by the user).

n/a

Because this breaks the API in an unavoidable way, it might be interesting to merge this before #828.


manifests/plugin/write_riemann.pp
```diff
@@ -1,25 +1,16 @@
...
-  Boolean $always_append_ds         = false,
-  Variant[Float,String] $ttl_factor = '2.0',
-  Boolean $check_thresholds         = false,
-  Array $tags                       = [],
-  Hash $attributes                  = {},
+  Array[Collectd::Write_riemann::Node] $nodes,
+  Enum['present', 'absent'] $ensure     = 'present',
+  Boolean $manage_package               = $collectd::manage_package,
+  Array[String[1]] $tags                = [],
+  Hash[String[1],String[1]] $attributes = {},
```
bastelfreak: more datatypes \o/

manifests/plugin/write_riemann.pp
```diff
@@ -28,6 +19,6 @@
 
   collectd::plugin { 'write_riemann':
     ensure  => $ensure,
-    content => template('collectd/plugin/write_riemann.conf.erb'),
+    content => epp('collectd/plugin/write_riemann.conf.epp'),
```
bastelfreak: epp \o/

bastelfreak:
Thanks for the awesome work @smortex !

PR #849, opened 2018-08-24
remove reference to collectd::plugin::perl::filename

mindriot88:
When using the perl plugin the following issue is raised during catalog compilation, upon inspection this appears to be declared from variable which is not used, so removing it seemed the most logic resolution.

`error during compilation: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Unknown variable: 'collectd::plugin::perl::filename'.`

### DIFF
--- a/manifests/plugin/perl/plugin.pp
+++ b/manifests/plugin/perl/plugin.pp
@@ -30,7 +30,6 @@ define collectd::plugin::perl::plugin (
   }
 
   $conf_dir = $collectd::plugin_conf_dir
-  $base_filename = $collectd::plugin::perl::filename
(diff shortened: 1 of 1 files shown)


bastelfreak:
Hi @mindriot88 , thanks for the fix! Can you add a unit test so we can ensure that it now works and compiles?

bastelfreak:
ping @mindriot88 :)

bastelfreak:
The varaible isn't used anywhere else:

```
bastelfreak@tmu ~/modulesync_config/modules/puppet-collectd $ git grep base_filename
bastelfreak@tmu ~/modulesync_config/modules/puppet-collectd $ 
```

PR #859, opened 2018-10-09
Persist the interval in the conf file as a number

nbarrientos:
When passing it as a string the plugin will basically ignore the setting and write the following to the log file:

```
  cf_util_get_cdtime:
     The Interval option requires exactly one numeric argument.
```

Without this patch it's not possible to declare different databases with different query interval.


traylenator:
This parameter is not currently tested but a good time to add it.

traylenator:
Would be good to convert this particular plugin to puppet 4 types as well. The whole module expects puppet 4.

nbarrientos:
I've added an extra assert to an existing test to make sure that the value ends up in the configuration file properly formatted.

alexjfisher:
I'm happy with other changes eg. to add data types and replace `create_resources` to be done in a separate PR.

alexjfisher:
@nbarrientos Thanks!

PR #871, opened 2019-03-13
add FlushInterval param to loadplugin config

zoojar:

### DIFF
--- a/manifests/plugin.pp
+++ b/manifests/plugin.pp
@@ -1,11 +1,12 @@
 #
 define collectd::plugin (
-  $ensure   = 'present',
-  $content  = undef,
-  $order    = '10',
-  $globals  = false,
-  $interval = undef,
-  $plugin   = $name
+  $ensure                          = 'present',
+  $content                         = undef,
+  $order                           = '10',
+  $globals                         = false,
+  $interval                        = undef,
+  $plugin                          = $name,
+  Optional[Integer] $flushinterval = undef,
 ) {
 
   include ::collectd
--- a/spec/acceptance/class_plugin_python_spec.rb
+++ b/spec/acceptance/class_plugin_python_spec.rb
@@ -57,7 +57,7 @@ describe 'collectd::plugin::python class' do
       # Run it twice and test for idempotency
       apply_manifest(pp, catch_failures: true)
       apply_manifest(pp, catch_changes: true)
-      shell('sleep 10')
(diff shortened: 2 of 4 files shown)


manifests/plugin.pp
```diff
@@ -1,11 +1,12 @@
...
-  $order    = '10',
-  $globals  = false,
-  $interval = undef,
-  $plugin   = $name
+  $ensure        = 'present',
+  $content       = undef,
+  $order         = '10',
+  $globals       = false,
+  $interval      = undef,
+  $flushinterval = undef,
```
bastelfreak: can you please add a datatype to the new parameter?

templates/loadplugin.conf.erb
```diff
@@ -2,6 +2,9 @@
 <% if scope.lookupvar('collectd::collectd_version_real') and (scope.function_versioncmp([scope.lookupvar('collectd::collectd_version_real'), '4.9.4']) >= 0) -%>
 <LoadPlugin <%= @plugin %>>
   Globals <%= @globals %>
+<% if @flushinterval and scope.lookupvar('collectd::collectd_version_real') and (scope.function_versioncmp([scope.lookupvar('collectd::collectd_version_real'), '5.2']) >= 0) -%>
+  FlushInterval <%= @flushinterval %>
```
bastelfreak: please add a unit test to validate that the condition works

zoojar: done!

manifests/plugin.pp
zoojar: done !

spec/acceptance/class_plugin_python_spec.rb
```diff
@@ -57,7 +57,7 @@
       # Run it twice and test for idempotency
       apply_manifest(pp, catch_failures: true)
       apply_manifest(pp, catch_changes: true)
-      shell('sleep 10')
+      shell('sleep 30')
```
bastelfreak: why did you need to increase this? Does collectd now start slower?

zoojar:
To try to get the tests to pass. Wondered if it was some kind of race
condition

spec/acceptance/class_plugin_python_spec.rb
zoojar: To try to get the tests to pass. Wondered if it was some kind of race
condition

PR #874, opened 2019-03-20
replace create_resources with resource types

### LINKED ISSUE #716: Replace create_resources() calls
`create_resources` calls should be replaced with proper loops

zoojar:
Replace create_resources with resource types

Fixes #716 (Replace create_resources() calls)

### DIFF
--- a/manifests/plugin/aggregation.pp
+++ b/manifests/plugin/aggregation.pp
@@ -16,5 +16,9 @@ class collectd::plugin::aggregation (
     'ensure' => $ensure,
   }
 
-  create_resources(collectd::plugin::aggregation::aggregator, $aggregators, $defaults)
+  $aggregators.each |String $resource, Hash $attributes| {
+    collectd::plugin::aggregation::aggregator { $resource:
+      * => $defaults + $attributes,
+    }
+  }
 }
--- a/manifests/plugin/postgresql.pp
+++ b/manifests/plugin/postgresql.pp
@@ -10,7 +10,10 @@ class collectd::plugin::postgresql (
 
   include ::collectd
 
-  $_manage_package = pick($manage_package, $::collectd::manage_package)
+  $_manage_package    = pick($manage_package, $::collectd::manage_package)
+  $databases_defaults = { 'ensure' => $ensure }
+  $queries_defaults   = { 'ensure' => $ensure }
+  $writers_defaults   = { 'ensure' => $ensure }
 
   if $facts['os']['family'] == 'RedHat' {
     if $_manage_package {
@@ -45,11 +48,21 @@ class collectd::plugin::postgresql (
     target  => "${collectd::plugin_conf_dir}/postgresql-config.conf",
   }
 
-  $defaults = {
-    'ensure' => $ensure,
+  $databases.each |String $resource, Hash $attributes| {
+    collectd::plugin::postgresql::database { $resource:
+      * => $databases_defaults + $attributes,
+    }
   }
 
-  create_resources(collectd::plugin::postgresql::database, $databases, $defaults)
-  create_resources(collectd::plugin::postgresql::query, $queries, $defaults)
-  create_resources(collectd::plugin::postgresql::writer, $writers, $defaults)
+  $queries.each |String $resource, Hash $attributes| {
+    collectd::plugin::postgresql::query { $resource:
+      * => $queries_defaults + $attributes,
+    }
+  }
+
+  $writers.each |String $resource, Hash $attributes| {
+    collectd::plugin::postgresql::writer { $resource:
+      * => $writers_defaults + $attributes,
+    }
+  }
 }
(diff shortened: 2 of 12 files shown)


juniorsysadmin:
Fixes #716

manifests/plugin/aggregation.pp
```diff
@@ -16,5 +16,10 @@
     'ensure' => $ensure,
   }
 
-  create_resources(collectd::plugin::aggregation::aggregator, $aggregators, $defaults)
+  $aggregators.each |String $resource, Hash $attributes| {
+    Resource['collectd::plugin::aggregation::aggregator'] {
```
alexjfisher: Would this form be more readable?
```puppet
collectd::plugin::aggregation::aggregator {$resource:
  * => $defaults + $attributes,
}
```

alexjfisher: Or do away with $defaults and then just
```puppet
collectd::plugin::aggregation::aggregator {$resource:
  ensure => $ensure,
  *      => $attributes,
}
```

bastelfreak: I agree with @alexjfisher here. I think his first option is shorter and easier to read.

manifests/plugin/powerdns.pp
```diff
@@ -3,13 +3,16 @@
   Enum['present', 'absent'] $ensure       = 'present',
   Integer $order                          = 10,
   Optional[Numeric] $interval             = undef,
-  Optional[Hash[String, Hash]] $servers   = undef,
-  Optional[Hash[String, Hash]] $recursors = undef,
+  Optional[Hash[String, Hash]] $servers   = {},
```
alexjfisher: Can we remove the `Optional`?

manifests/plugin/postgresql.pp
```diff
@@ -10,7 +10,10 @@
 
   include ::collectd
 
-  $_manage_package = pick($manage_package, $::collectd::manage_package)
+  $_manage_package    = pick($manage_package, $::collectd::manage_package)
+  $databases_defaults = { 'ensure' => $ensure }
+  $queries_defaults   = { 'ensure' => $ensure }
+  $writers_defaults   = { 'ensure' => $ensure }
```
alexjfisher: Why create all these identical hashes?

@bastelfreak Did you not like https://github.com/voxpupuli/puppet-collectd/pull/874#discussion_r267584362 ?

manifests/plugin/powerdns.pp
zoojar: yeah sure,  i didn't want to change too much! just doing the bare minimum to get create_resources replaced

manifests/plugin/postgresql.pp
zoojar: looool i thought this too. It;'s one way of separating defaults for each resource

zoojar:
```
    collectd::plugin::postgresql::writer { $resource:
      * => { 'ensure' => $ensure } + $attributes,
    }
```
better?

alexjfisher: With `create_resources` we had to pass a hash for the defaults (although it didn't strictly need to be assigned to a variable first).

Since we just have a single default property (`ensure`), putting it directly in the resource declaration is probably cleaner??

manifests/plugin/aggregation.pp
zoojar: > Or do away with $defaults and then just
> 
> ```puppet
> collectd::plugin::aggregation::aggregator {$resource:
>   ensure => $ensure,
>   *      => $attributes,
> }
> ```

this one won't merge, so we are assuming that attributes doesn't contain ensure

manifests/plugin/postgresql.pp
alexjfisher: > ```
>     collectd::plugin::postgresql::writer { $resource:
>       * => { 'ensure' => $ensure } + $attributes,
>     }
> ```
> better?

No need to merge hashes.  Just...
```puppet
collectd::plugin::postgresql::writer { $resource:
  ensure => $ensure,
  *      => $attributes,
}
```

alexjfisher: That depends on `$attributes` not containing an override for `ensure` though. (which might be a perfectly reasonable assumption?)

zoojar: they do contain them....
https://github.com/voxpupuli/puppet-collectd/blob/master/manifests/plugin/postgresql/writer.pp#L3

alexjfisher: Actually... I don't think that is a safe assumption.
` * => { 'ensure' => $ensure } + $attributes,` works for me.

zoojar: ok. well, we still get one thing... replacement of create_resources :)

manifests/plugin/powerdns.pp
zoojar: @alexjfisher  removed

PR #875, opened 2019-03-22
puppet-lint: fix topscope_variable

bastelfreak:


binford2k:
works for me, but let's see what @mmerfort thinks

PR #880, opened 2019-05-29
plugin/java: handle OpenJDK as well

GiedriusS:
`collectd-java` requires:
* `libjvm.so()(64bit)`
* `libjvm.so(SUNWprivate_1.1)(64bit)`

On CentOS/RHEL 6.x/7.x these are provided by OpenJDK. OpenJDK stores the
JVM files in a bit of a different location: `amd64` has been removed
from the path. Thus, ATM the link does not get created properly on such
systems.

Fix this problem by adding two `exec` resources which both will try to
create the needed libjvm symbolic link. They are both smart and check if
that file exists before trying to execute. An assumption is added to the
code that both of the files cannot exist at the same time.

### DIFF
--- a/manifests/plugin/java.pp
+++ b/manifests/plugin/java.pp
@@ -36,6 +36,7 @@ class collectd::plugin::java (
       # Reload SO files so libjvm.so can be found
       exec { '/sbin/ldconfig':
         unless  => '/sbin/ldconfig -p |grep libjvm.so >/dev/null 2>&1',
+        require => [Exec['Link libjvm.so on OpenJDK'], Exec['Link libjvm.so on Oracle']]
       }
     }
   }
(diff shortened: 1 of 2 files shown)


manifests/plugin/java.pp
```diff
@@ -19,13 +19,23 @@
       }
     }
     if $java_home {
-      file { '/usr/lib64/libjvm.so':
-        ensure => 'link',
-        target => "${java_home}/jre/lib/amd64/server/libjvm.so",
+      # OpenJDK based Java distribution
+      exec { "/usr/bin/ln -s /usr/lib64/libjvm.so ${java_home}/jre/lib/server/libjvm.so":
```
bastelfreak: Why did you replace the file resource with an exec one? The file resource can handle symlinks.

GiedriusS: Indeed it can and I thought about it but how would you implement a conditional `File` resource which would create `/usr/lib64/libjvm.so` depending on certain conditions i.e. whether `${java_home}/jre/lib/server/libjvm.so` exists or `${java_home/jre/lib/amd64/server/libjvm.so`? Maybe I am missing something here but the `File` resource doesn't have an `onlyif` parameter and the only way AFAICT to do something like this is via an `exec` resource.

bastelfreak:
Thanks for the fix!

PR #882, opened 2019-06-12
Add puppet tasks to call collectdctl

traylenator:

Two puppet tasks

* collectd::listval - lists all collectd metrics on a node
* collectd::getval metric=load/load-relative - shows values
  of a particular metric

For example

```bash
$ bolt -u root task run collectd::getval metric=load/load-relative -n node.example.org
Started on node.example.org...
Finished on node.example.org:
  {
    "metric": "load/load-relative",
    "values": {
      "shortterm": "1.750000e-01",
      "longterm": "8.000000e-02",
      "midterm": "8.500000e-02"
    }
  }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -2092,6 +2092,57 @@ collectd::typesdb { $db:
 }
 ```
 
+## Puppet Tasks
+Assuming that the collectdctl command is available on remote nodes puppet tasks exist to
+run collectdctl and collect results from nodes.
+
+### Puppet Task collectd::listval
+```bash
+$ bolt task show collectd::listval
+collectd::listval - Lists all available collectd metrics
+
+USAGE:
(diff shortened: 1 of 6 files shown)


traylenator:
I added a new dependency to metadata.yaml - I'm easily persuaded this should just be a comment in the README.

traylenator:
There is a massive lack of tests here. Very happy to do some but we need some framework.

alexjfisher:
> I added a new dependency to metadata.yaml - I'm easily persuaded this should just be a comment in the README.

I think it's fine in the metadata.

traylenator:
It's broken on centos6 with python 2.6 . I have a fix migrating to `python3` which is available on 6 also.
Another commit here?

PR #888, opened 2019-08-02
Plugin write kafka custom properties and meta

nitrik:
Allow custom Property and Meta.
Custom properties would allow, for example, sasl_ssl configuration to be added to the plugin.
Custom metas would allow specific parameters.

No fix, new feature.

### DIFF
--- a/manifests/plugin/write_kafka.pp
+++ b/manifests/plugin/write_kafka.pp
@@ -4,6 +4,8 @@ class collectd::plugin::write_kafka (
   Array[String] $kafka_hosts = ['localhost:9092'],
   $kafka_port                = 9092,
   Hash $topics               = {},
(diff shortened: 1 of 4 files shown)


manifests/plugin/write_kafka.pp
```diff
@@ -4,9 +4,11 @@
   Array[String] $kafka_hosts = ['localhost:9092'],
   $kafka_port                = 9092,
   Hash $topics               = {},
+  Hash $properties           = {},
+  Hash $meta                 = {},
 ) {
 
-  include collectd
+  include ::collectd
```
bastelfreak: Please don't add the topscope to the include. That's a deprecated puppet 3 syntax.

PR #889, opened 2019-08-06
Apt update before installation of collectd

theosotr:
Hi

It seems to me that there is not any dependency between Exec['apt_update'] and Package['collectd'].
Specifically, this is what I get in `noop` mode: `apt-update` is run after the installation of the `collectd` package.

```
Notice: Compiled catalog for 067d02440e1f in environment production in 0.48 seconds
Notice: /Stage[main]/Apt/Apt::Setting[conf-update-stamp]/File[/etc/apt/apt.conf.d/15update-stamp]/ensure: current_value absent, should be file (noop)
Notice: Apt::Setting[conf-update-stamp]: Would have triggered 'refresh' from 1 events
Notice: /Stage[main]/Apt/File[preferences]/ensure: current_value absent, should be file (noop)
Notice: Class[Apt]: Would have triggered 'refresh' from 2 events
Notice: Class[Collectd::Repo::Debian]: Would have triggered 'refresh' from 1 events
Notice: /Stage[main]/Collectd::Install/Package[collectd]/ensure: current_value purged, should be present (noop)
Notice: /Stage[main]/Collectd::Install/Package[collectd-core]/ensure: current_value purged, should be present (noop)
Notice: Class[Collectd::Install]: Would have triggered 'refresh' from 2 events
Notice: Class[Apt::Update]: Would have triggered 'refresh' from 2 events
Notice: /Stage[main]/Apt::Update/Exec[apt_update]: Would have triggered 'refresh' from 1 events
Notice: Class[Apt::Update]: Would have triggered 'refresh' from 1 events
Notice: /Stage[main]/Collectd::Config/File[collectd.d]/ensure: current_value absent, should be directory (noop)
```
You can reproduce it by viewing the compiled catalog.

Indeed, the class `Apt::Update` is contained inside `Stage[main]` and not inside the class `Collectd::Repo::Debian` as you might expect.

### DIFF
--- a/manifests/repo/debian.pp
+++ b/manifests/repo/debian.pp
@@ -2,7 +2,7 @@ class collectd::repo::debian {
 
   contain apt
   if $collectd::manage_package {
-    Exec['apt_update'] -> Package[$collectd::package_name]
+    Class['apt::update'] -> Package[$collectd::package_name]
   }
 
   if $collectd::ci_package_repo {
--- a/manifests/repo/debian.pp
+++ b/manifests/repo/debian.pp
@@ -1,6 +1,9 @@
 class collectd::repo::debian {
 
   contain apt
+  if $collectd::manage_package {
+    Exec['apt_update'] -> Package[$collectd::package_name]
+  }
 
   if $collectd::ci_package_repo {
 
--- a/spec/classes/collectd_init_spec.rb
+++ b/spec/classes/collectd_init_spec.rb
@@ -192,6 +192,32 @@ describe 'collectd', type: :class do
             end
           end
 
+          context 'and manage_package is true' do
+            let(:params) do
+              {
+                manage_repo: true,
+                manage_package: true
+              }
+            end
+
+            if facts[:osfamily] == 'Debian'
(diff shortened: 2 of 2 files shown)


bastelfreak:
Hi @theosotr, thanks for the PR. Can you please add  a test for this change?

theosotr:
@bastelfreak, thanks for your reply.

I added the test cases as you requested.

theosotr:
More details in the official documentation of `puppetlabs-apt`:
https://github.com/puppetlabs/puppetlabs-apt#adding-new-sources-or-ppas

bastelfreak:
@theosotr can you take a look at the failing spec tests?

theosotr:
@bastelfreak it seems to me that the error is not related to my patch.

```
..

An error occurred while loading ./spec/defines/collectd_plugin_curl_json_spec.rb.

Failure/Error: case facts[:os]['family']

NoMethodError:

  undefined method `[]' for nil:NilClass
```

bastelfreak:
@theosotr sorry that this took so long. I fixed the error in another PR and rebased this one. Will merge it when it turns green.

PR #892, opened 2019-10-03
Add ReconnectInterval option to the write_graphite plugin (defaults to zero, introduced in version 5.6)

markasammut:

Version 5.6 has introduced the ReconnectInterval option to the write_graphite plugin (https://collectd.org/wiki/index.php/Version_5.6).

ReconnectInterval Seconds
When set to non-zero, forces the connection to the Graphite backend to be closed and re-opend periodically. This behavior is desirable in environments where the connection to the Graphite backend is done through load balancers, for example. When set to zero, the default, the connetion is kept open for as long as possible.
(https://collectd.org/documentation/manpages/collectd.conf.5.shtml#plugin_write_graphite)

New feature

### DIFF
--- a/manifests/plugin/write_graphite/carbon.pp
+++ b/manifests/plugin/write_graphite/carbon.pp
@@ -12,6 +12,7 @@ define collectd::plugin::write_graphite::carbon (
   $protocol                  = 'tcp',
   Boolean $separateinstances = false,
   Boolean $logsenderrors     = true,
+  $reconnectinterval         = 0,
 ) {
 
   include collectd
--- a/manifests/plugin/write_graphite/carbon.pp
+++ b/manifests/plugin/write_graphite/carbon.pp
@@ -12,7 +12,7 @@ define collectd::plugin::write_graphite::carbon (
   $protocol                  = 'tcp',
   Boolean $separateinstances = false,
   Boolean $logsenderrors     = true,
-  $reconnectinterval         = 0,
+  Integer $reconnectinterval = 0,
 ) {
 
   include collectd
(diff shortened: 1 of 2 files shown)


templates/plugin/write_graphite/carbon.conf.erb
```diff
@@ -17,7 +17,10 @@
   LogSendErrors <%= @logsenderrors %>
   Protocol "<%= @protocol %>"
 <%- end -%>
-<%- if scope.lookupvar('collectd::collectd_version_real') and (scope.function_versioncmp([scope.lookupvar('collectd::collectd_version_real'), '5.3']) >= 0) -%>
```
igalic: if this was an EPP template, these `scope.lookupvar()` calls could simply be a `$`

markasammut: I know, but for consistency I kept the same format as the rest of the template (https://github.com/voxpupuli/puppet-collectd/blob/master/templates/plugin/write_graphite/carbon.conf.erb#L16 and https://github.com/voxpupuli/puppet-collectd/blob/master/templates/plugin/write_graphite/carbon.conf.erb#L20)

igalic: aye, hence why this is just a comment, not a request for you to change anything.

let's do this in a follow up pr, tho.

manifests/plugin/write_graphite/carbon.pp
```diff
@@ -12,6 +12,7 @@
   $protocol                  = 'tcp',
   Boolean $separateinstances = false,
   Boolean $logsenderrors     = true,
+  $reconnectinterval         = 0,
```
bastelfreak: Can you please add a suitable datatype for this parameter?

markasammut: Done, thanks

PR #896, opened 2019-11-18
RegexProcess has been renamed during collectd

### LINKED ISSUE #897: ProcEvent plugin uses ProcessRegex instead of RegexProcess
The config created by puppet-collectd uses RegexProcess, see https://github.com/voxpupuli/puppet-collectd/blob/master/templates/plugin/procevent.conf.epp#L6

Collectd uses ProcessRegex. This is fixed by https://github.com/voxpupuli/puppet-collectd/pull/896

mrunge:
review. It is now named ProcessRegex.

See https://github.com/collectd/collectd/blob/master/src/procevent.c#L1227

Fixes #897

### DIFF
--- a/manifests/plugin/procevent.pp
+++ b/manifests/plugin/procevent.pp
@@ -20,7 +20,7 @@
 #  name will be monitored for EXECs and EXITs.
 #  Defaults to undef
 #
-# [*regex_process*]
+# [*process_regex*]
 #  Enumerate a process pattern to monitor.  All processes that match this
 #  regular expression will be monitored for EXECs and EXITs.
 #  Defaults to undef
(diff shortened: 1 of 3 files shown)


mrunge:
Yay, CI is friendly to me.

Since the name of the parameter in collectd has been renamed, and the plugin is included in collectd-5.10, this should be fine.

mrunge:
Here is the thing, only *this* config is valid, the other one was only accepted in early versions of the plugin in collectd. The plugin itself was only recently merged into collectd-5.10. 

The only possibility to have collectd accepting the currently implemented config (from puppet-collectd), you'd have to cherry-pick an early development snapshot of the plugin (more than a year old) and build collectd yourself. That is nothing, a regular user would do. 

The other way around: chances are pretty high, that collectd won't accept the currently created config.

Does that make sense?

igalic:
yes

mrunge:
Thank you!

PR #899, opened 2019-12-06
modulesync 2.10.1 / Drop FreeBSD 9 & 10 / Add FreeBSD 11 and 12 / Drop Solaris

dhoppe:


bastelfreak:
I had to remove some of the legacy operating systems. They were all end of life and facterdb doesn't ship recent facts for it.

PR #906, opened 2020-02-08
Change all `port` parameters to use `Stdlib::Port`

bastelfreak:
Instead of having a mix of strings and integers, all port parameters
should be using `Stdlib::Port`.

I've also changed some `String`s to `Stdlib::Host`, but none of these
should be breaking changes.

### DIFF
--- a/manifests/plugin/ovs_stats.pp
+++ b/manifests/plugin/ovs_stats.pp
@@ -27,7 +27,7 @@
 #  Defaults to 'collectd-ovs_stats'
 #
 # [*port*]
-#  TCP-port to connect to. Either a service name or a port number may be given.
+#  TCP-port to connect to.
 #
 # [*socket*]
 #  The UNIX domain socket path of OVS DB server JSON-RPC interface used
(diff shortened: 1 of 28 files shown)


bastelfreak:
Replaces https://github.com/voxpupuli/puppet-collectd/pull/870

I created a new PR because I wasn't allowed to push to your branch.

bastelfreak:
Hey people. Would be nice if someone can review and approve this. I would like to wait with the merge until we've some other backwards incompatible changes.

manifests/plugin/ovs_stats.pp
```diff
@@ -27,7 +27,7 @@
 #  Defaults to 'collectd-ovs_stats'
 #
 # [*port*]
-#  TCP-port to connect to. Either a service name or a port number may be given.
+#  TCP-port to connect to.
```
igalic: so this pr is compatibility break

manifests/plugin/postgresql/database.pp
```diff
@@ -2,9 +2,9 @@
 # useful if you have multiple instances of different version of pg
 define collectd::plugin::postgresql::database (
   $ensure       = 'present',
```
igalic: this could use more types ;)

PR #909, opened 2020-02-12
Plugin Mcelog default config

MichalRebisz:
Plugin mcelog default config.


bastelfreak:
@MichalRebisz do you know what the default values form collectd are for those settings? Does this break existing configs for people?

MichalRebisz:
> @MichalRebisz do you know what the default values form collectd are for those settings? Does this break existing configs for people? Also please suqash the commits down to one.

Thous are the values I received as default for mcelog from the collectd team. Whether this breaks exsiting configs for people well it shouldn't since mcelog was added about a week ago but I did not have the default values at that point.

bastelfreak:
thanks for the explanation!

PR #911, opened 2020-02-13
Add pcie_errors plugin

prabiegx:
The pcie_errors plugin collects PCI Express errors from Device Status in Capability structure and from Advanced Error Reporting Extended Capability where available.


manifests/plugin/pcie_errors.pp
```diff
@@ -0,0 +1,15 @@
+class collectd::plugin::pcie_errors (
```
bastelfreak: can you please add puppet-string documentation for the new parameters?

manifests/plugin/pcie_errors.pp
```diff
@@ -0,0 +1,15 @@
...
+  Enum['present', 'absent'] $ensure                   = 'present',
+  Optional[String]          $source                   = 'sysfs',
+  Optional[String]          $access_dir               = undef,
+  Optional[Boolean]         $report_masked            = false,
+  Optional[Boolean]         $persistent_notifications = false,
```
bastelfreak: does Optional make sense here? In most cases undef isn't a useful value for a Boolean variable.

templates/plugin/pcie_errors.conf.epp
```diff
@@ -0,0 +1,14 @@
+<Plugin pcie_errors>
```
bastelfreak: please add a comment like `# THIS FILE IS MANAGED BY PUPPET`

prabiegx: "# Generated by Puppet" line is already generated by Puppet. Test case for default values ensures that this line is present.

manifests/plugin/pcie_errors.pp
prabiegx: I added documentation for the parameters and changed Boolean ones to not be optional.

manifests/plugin/pcie_errors.pp
```diff
@@ -0,0 +1,23 @@
...
+#
+# @param ensure Ensure param for collectd::plugin type.
+# @param source Use sysfs or proc to read data from /sysfs or /proc.
+# @param access_dir Directory used to access device config space.
+# @param report_masked If true plugin will notify about errors that are set to masked in Error Mask register.
+# @param persistent_notifications If false plugin will dispatch notification only on set/clear of error.
+#
+class collectd::plugin::pcie_errors (
+  Enum['present', 'absent'] $ensure                   = 'present',
+  Optional[String]          $source                   = 'sysfs',
```
bastelfreak: We are never able to set $source to undef if it has a default value. We can only change `sysfs`  to another string. Are sysfs and proc the only valid values? In that case we could use an enum here?

templates/plugin/pcie_errors.conf.epp
bastelfreak: ah right, stupid me!

manifests/plugin/pcie_errors.pp
prabiegx: Good point. Values other than sysfs and proc are treated as error. As suggested I changed data type to enum and updated test case.

prabiegx:
It looks like build failed due to one of the configured repositories failing, which seems to be "random" issue not related to this change. @bastelfreak could failing build be repeated?

PR #912, opened 2020-02-14
Add Plugin Log parser

MichalRebisz:
Added log parser plugin.


templates/plugin/log_parser.conf.epp
```diff
@@ -0,0 +1,70 @@
+<%- | Optional $logfile = undef | -%>
```
ghoneycutt: This should have a spec test that shows how the template would be rendered given the default setting. Check out an example at https://github.com/ghoneycutt/puppet-module-bind/blob/master/spec/defines/acl_spec.rb#L22-L47

types/log_parser/logfile.pp
```diff
@@ -0,0 +1,5 @@
+#https://wiki.opnfv.org/display/fastpath/Logparser+plugin+HLD
```
ghoneycutt: The types should have spec tests. Here's an example at https://github.com/sensu/sensu-puppet/blob/master/spec/type_aliases/backend_url_spec.rb

Basically you have two arrays, one of things that should pass and another of things that should fail.

templates/plugin/log_parser.conf.epp
bastelfreak: can you please add a statement like `# THIS FILE IS MANAGED BY PUPPET`?

types/log_parser/logfile.pp
MichalRebisz: Should I use the same approach? Asking since I see it uses shoulda-matchers (If im not mistaken allow_value method is a part of it) Asking since if i'm to use the same approach to testing it I would have to add shoulda-matchers since I dont believe that puppet uses it right now. Or I could go with the approach that is currently example at https://github.com/voxpupuli/puppet-collectd/blob/e1f29e95d3af724eb25b752b16594463c282e7e1/spec/defines/collectd_plugin_snmp_data_spec.rb but that only test if the values are set if im not wrong.

templates/plugin/log_parser.conf.epp
bastelfreak: ups, that isn't required

spec/classes/collectd_plugin_log_parser_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+                'firstfullread' => false
+              }
+            ]
+          }
+        end
+
+        it "Will create #{options[:plugin_conf_dir]}/06-log_parser.conf" do
+          is_expected.to contain_file('log_parser.load').with(
+            ensure: 'present',
+            path: "#{options[:plugin_conf_dir]}/06-log_parser.conf"
```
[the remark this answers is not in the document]
MichalRebisz: Is this what you had in mind?

bastelfreak:
The test looks good. Sadly something changed within the puppet 5 debian package or the image and it stopped working. This effects not only our collectd module but also all other modules we have. Hopefully Puppet can fix this during the next week.

MichalRebisz:
> The test looks good. Sadly something changed within the puppet 5 debian package or the image and it stopped working. This effects not only our collectd module but also all other modules we have. Hopefully Puppet can fix this during the next week.

So do we have to wait for a fix for puppet or could we get this PR merged before that?

bastelfreak:
We need to wait until it's fixed.hopefully it's fixed soon

PR #913, opened 2020-02-18
Add dpdk_telemetry plugin

prabiegx:
The dpdk_telemetry plugin collects DPDK ethernet device metrics via dpdk_telemetry library.


manifests/plugin/dpdk_telemetry.pp
```diff
@@ -0,0 +1,28 @@
...
+# delivering the requested metrics. Plugin parses the JSON data
+# and publishes the metric values to collectd for further use.
+#
+# @param ensure Ensure param for collectd::plugin type.
+# @param client_socket_path UNIX domain client socket to receive messages from DPDK telemetry library.
+# @param dpdk_socket_path UNIX domain DPDK telemetry socket to be connected to send messages.
+#
+class collectd::plugin::dpdk_telemetry (
+  Enum['present', 'absent'] $ensure             = 'present',
+  Optional[String]          $client_socket_path = '/var/run/.client',
```
ghoneycutt: Use `Optional[Stdlib::Absolutepath]` instead as this value is either undef or a path to a file.

spec/classes/collectd_plugin_dpdk_telemetry_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+          is_expected.to compile.with_all_deps
+          is_expected.to contain_file('dpdk_telemetry.load').with(
+            ensure: 'present',
+            path: "#{options[:plugin_conf_dir]}/10-dpdk_telemetry.conf",
+            content: %r{ClientSocketPath "/test/path/.client"}m
+          )
+        end
+      end
+
+      context ':ensure => present and :dpdk_socket_path => /test/path/telemetry' do
```
ghoneycutt: add a context for `ensure => absent`

spec/classes/collectd_plugin_dpdk_telemetry_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+        let :params do
+          { dpdk_socket_path: '/test/path/telemetry' }
+        end
+
+        it "Will create #{options[:plugin_conf_dir]}/10-dpdk_telemetry.conf" do
```
ghoneycutt: Since `client_socket_path` can be `undef`, there should be a test that shows the content when the param is `undef`. Same for `dpdk_socket_path`.

ghoneycutt:
Thanks for the great contribution!

manifests/plugin/dpdk_telemetry.pp
bastelfreak: if it's an `Optional[]` datatype, the default value has to be `undef`. If you want to pass a default path, like `/var/run/.client`, it doesn't need to be `Optional[]` because you cannot pass `undef` to it. Same applies for `$dpdk_socket_path`

ghoneycutt: The default value could be, but does not have to be `undef`. Since the value is either `undef` or a path to a file, then `Stdlib::Absolutepath` makes more sense than `String`.

prabiegx: I changed String to Stdlib:Absolutepath and kept "Optional[...]" Is that all right?

spec/classes/collectd_plugin_dpdk_telemetry_spec.rb
prabiegx: Test case with "ensure => absent" added.

spec/classes/collectd_plugin_dpdk_telemetry_spec.rb
prabiegx: I was looking into creating test case that you described but I stumbled upon obstacle. I wrote following test:

```
      context ':ensure => present, :client_socket_path => undef and :dpdk_socket_path => undef' do
        let :params do
          { client_socket_path: :undef,
            dpdk_socket_path: :undef }
        end

        sth = <<EOS
# Generated by Puppet
<LoadPlugin dpdk_telemetry>
  Globals false
</LoadPlugin>

<Plugin dpdk_telemetry>
</Plugin>

EOS

        it "Will create #{options[:plugin_conf_dir]}/10-dpdk_telemetry.conf" do
          is_expected.to compile.with_all_deps
          is_expected.to contain_file('dpdk_telemetry.load').with(
            ensure: 'present',
            path: "#{options[:plugin_conf_dir]}/10-dpdk_telemetry.conf",
            content: sth
          )
        end
      end
```
	 
Configuration is generated using specified default values rather than assuming that no value was provided.
I found interesting message by _jcbollinger_ [here](https://groups.google.com/d/msg/puppet-users/6nL2eROH8is/Fg6W5lEjHQsJ), suggesting that assigning undef doesn't actually assign undef, only implies that no specified value is being assigned, and therefore default should be used.

I'm relatively new to Puppet, RSpec and Ruby so I can't be sure, however my observations suggests that it in fact might work that way. Additionally, I couldn't find this kind of test case in this repository. Are you sure that test case you described is possible to write? If so, could you give any suggestion how to proceed?

manifests/plugin/dpdk_telemetry.pp
bastelfreak: @ghoneycutt if you pass undef to a variable that has a default value, undef won't overwrite it. In my opinion the `Optional[]` needs to be removed.

prabiegx: I removed `Optional[]` since, as @bastelfreak  mentioned, we can't change it back to `undef` if we already have defined default value. Without `Optional[]` there is now no need for `*_socket_path => undef` test case.

prabiegx: @ghoneycutt  are you satisfied with applied modifications?

prabiegx:
Changes squashed and rebased.

prabiegx:
Is this ready for merge or needs further changes?

bastelfreak:
Thanks for the PR @prabiegx !

PR #914, opened 2020-02-20
Add dcpmm plugin

prabiegx:
The dcpmm plugin will collect Intel(R) Optane(TM) DC Persistent Memory related performance statistics.


prabiegx:
One of the CI builds did not pass, but issue looks to be result of unstable build environment and not related to this change. @bastelfreak can this job be re-run?

PR #917, opened 2020-03-06
Add CentOS 8 support

traylenator:

Add CentOS 8 support

The main differences for CentOS 8 are:
* There is now a new collectd-utils package containing collectdctl
* collectd-python is now a seperate sub package
* python is python3.
* pip command is now pip3
* The rabbitmq plugin no longer works due to lack of python3 support

The `python_dir` fact will fall back to `/usr/libexec/platform-python`
if `python3` does not exist on the path. This is the python
that collectd uses and  significantly all that collectd-python requires.

Note that setting `collectd::ci_package_repo` true will currently
result in yum being configured with a non existant repo since
https://pkg.ci.collectd.org/rpm/ does not exist
for 8.  [request](https://github.com/collectd/collectd-ci/issues/35)

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -4,6 +4,7 @@
     - set: ubuntu1604-64
     - set: ubuntu1804-64
     - set: centos7-64
+    - set: centos8-64
     - set: debian8-64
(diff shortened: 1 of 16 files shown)


traylenator:
I think after this it makes sense to add a `Boolean collectd::utils = false` param that ensures
collectdctl is present installing the `collectd-utils` package depending on the OS.

bastelfreak:
@traylenator thanks for the awesome work! Is this still WIP?

traylenator:
One more feature to add I'll make a release hopefully.

PR #919, opened 2020-03-10
New utils parameter to install collectdctl

traylenator:
On some operating systems `collectdctl`, `collectd-nagios`
is contained in a seperate sub package, typically `collectd-utils`

Setting `utils` to `true` will install this sub package if it exists
for an operating system.

Also start some puppet-strings.


bastelfreak:
thanks for the PR!

PR #920, opened 2020-03-10
write_http/disk is own sub package on CentOS 8

traylenator:
####  Pull Request (PR) description
As of CentOS 8 the write_http and disk plugins are  in their own subpackage
so must be installed on CentOS 8 in these plugins are enabled.

Note this also contains a bugfix for CentOS 7. On CentOS 7 at least as far
as EPEL7 collectd is concerned there is no collectd-disk package and so the attempt on
CentOS 7 to install this `collectd-disk` (by default) is no longer made.

acceptance tests for plugins have been added.


traylenator:
WIP until https://github.com/voxpupuli/puppet-collectd/pull/919 is merge.

bastelfreak:
@traylenator do you want/need to rebase here?

manifests/plugin/write_http.pp
```diff
@@ -1,8 +1,14 @@
...
+# @see  https://collectd.org/wiki/index.php/Plugin:Write_HTTP
+#
+# @parameter Manage a collectd-write_http package? If undef a suitable value per OS will be chosen.
+#
 class collectd::plugin::write_http (
   Enum['present', 'absent']          $ensure = 'present',
   Hash[String, Hash[String, Scalar]] $nodes = {},
-  Hash[String, Hash[String, Scalar]] $urls  = {}
+  Hash[String, Hash[String, Scalar]] $urls  = {},
+  Optional[Boolean]                  $manage_package = undef,
```
bastelfreak: I'm not a big fan of an Optional Boolean. Can't we set the default to true or fals?

manifests/plugin/write_http.pp
```diff
@@ -11,6 +17,21 @@
     fail('Only one of nodes or urls is supposed to be defined')
   }
 
+  if $manage_package !~ Undef {
+    $_manage_package = $manage_package
+  } else {
+    if $facts['os']['family'] == 'RedHat' and versioncmp($facts['os']['release']['major'],'8') >= 0 {
```
bastelfreak: We could get rid of the Optional if this logic is in the params.pp and $manage_packages uses a variable from params.pp?

manifests/plugin/write_http.pp
traylenator: In this case there are 3 states. 
* Use the default for operating system (so false on debian, true on redhat 8)
* Force to be false.
* Force to be true.

manifests/plugin/write_http.pp
traylenator: Yes that would work however currently none of the per plugin class inherit from init, params.

We would explode the number of parameters on init.pp if we went down that road.

Would make a lot of sense to switch to module level hiera to achieve this if it were done.

traylenator:
Rebased.

manifests/plugin/write_http.pp
traylenator: Basically there is a matrix of OS vs plugins which you need to keep somewhere.

The generall direction of travel is that RedHat is splitting package while Debian remains as one package so it would at least become a lot simpler in the future so adding 
plugin setting per package now would be a problem for the future.

bastelfreak: Did you try to access a variable from params.pp? each plugin includes collectd and that inherits from params so this should work?

traylenator: I think it will work, I am just questioning if it is a good idea.

We would need to use `collectd::plugin_http_write_manage_package` and another
one for every plugin, or least where is needs to change based on OS.

And if we go further `collectd::plugin_systemd_package_provide` to cope where
we have rpms on RedHat and pips on debian.

traylenator: One more thing to add.
Such a change would be a major rework of the module, certainly outside the scope of this
bug fix. The `undef/true/false` system is used all over the place currently.


Shall I  create an issue. **Move all OS logic out to module level hiera data** 
With module hiera data we could just define default values for 
`collectd::plugin::http_write` per OS and not pollute init with per metric parameters.

traylenator: And a sub question.

What's the preference for the case where a parameter has per OS defaults in hiera data?
Just define in class as 
```puppet
Boolean $manage_package,
```
with no default

bastelfreak: my opinion here:
* Data that's the same across all operating systems should not be in module level hiera nor params.pp. It should be directly in the related class
* Everything thats dynamic and can be calcuated within data in modules should go there.

It would be awesome if you could raise an Issue/PR for data in modules. If the data is coming from data in modules the parameter should have no default within the class.

traylenator: Issue created https://github.com/voxpupuli/puppet-collectd/issues/921

Will try some time to look, definitely a worth while task.

@bastelfreak thanks for comments/review.

PR #918, opened 2020-03-09
Require puppet-epel over stahnma-epel

traylenator:
#### Pull Request (PR) description

stahnma-epel has now migrated to VoxPupuli
name space.

Also in particular the `puppet-epel` version is requirment to
work with CentOS 8.

PR #922, opened 2020-03-16
Release 11.3.0

traylenator:
Release 11.3.0

Introduces first CentOS 8 support in particular.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,26 @@ All notable changes to this project will be documented in this file.
 Each new release typically also includes the latest modulesync defaults.
 These should not affect the functionality of the module.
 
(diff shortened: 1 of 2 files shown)


bastelfreak:
Before we merge this: I pinged in https://github.com/voxpupuli/puppet-collectd/pull/918 . I'm not sure if this qualifies as backwards-incompatible. if it is, we should also include https://github.com/voxpupuli/puppet-collectd/pull/906. But I vote for a minor release, not major.

bastelfreak:
thanks for the PR, I did the travis_release rake task

traylenator:
@bastelfreak thanks - I might actually have some monitoring on my new services now  - winging it.

Issue #901, opened 2019-12-26
Error with Plugin "Load" with collectd-5.9 and RHEL-8


dusharu:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 5.5.17
- Ruby:
- Distribution: Red Hat Enterprise Linux release 8.1 (Ootpa)
- Module version: puppet-collectd (v10.1.0)
- Collectd: 5.9.0

## How to reproduce (e.g Puppet code you use)
```
class { 'collectd::plugin::load': }
```

## What are you seeing

Try test config after puppet run.
```
# collectd -T
Error: Parsing the config file failed!
```

## What behaviour did you expect instead
Disable template load.conf.erb for collectd 5.9
https://github.com/voxpupuli/puppet-collectd/blob/master/templates/plugin/load.conf.erb

It works after comment `<Plugin load>` sections in config.
```
# vim /etc/collectd.d/10-load.conf
###### /etc/collectd.d/10-load.conf ###### 
# Generated by Puppet
<LoadPlugin load>
  Globals false
</LoadPlugin>

#<Plugin load>
#  ReportRelative false
#</Plugin>
###### /etc/collectd.d/10-load.conf ###### 

# collectd -T
# systemctl restart collectd
# systemctl status collectd |grep ac
   Active: active (running) since Thu 2019-12-26 16:12:41 MSK; 6s ago
```

## Output log
```
# puppet agent -t
Info: Using configured environment 'develop'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Loading facts
Info: Applying configuration version '1577365409'
Info: Computing checksum on file /etc/collectd.d/.10-load.conf.swp
Info: /Stage[main]/Collectd::Config/File[/etc/collectd.d/.10-load.conf.swp]: Filebucketed /etc/collectd.d/.10-load.conf.swp to puppet with sum fc107359032900b1f7da7d76b871eeb8
Notice: /Stage[main]/Collectd::Config/File[/etc/collectd.d/.10-load.conf.swp]/ensure: removed (corrective)
Info: Class[Collectd::Config]: Scheduling refresh of Class[Collectd::Service]
Info: Class[Collectd::Service]: Scheduling refresh of Service[collectd]
Notice: /Stage[main]/Qas/Exec[vastool_timesync]/returns: executed successfully (corrective)
Notice: /Stage[main]/Collectd::Plugin::Load/Collectd::Plugin[load]/File[load.load]/content: 
[...]

PR #924, opened 2020-03-20
Skip Load:ReportRelative as problematic for collectd 5.9.0

traylenator:

Re-fixes https://github.com/voxpupuli/puppet-collectd/issues/901

The bug was that a ReportRelative configuration to the load
plugin for collectd 5.9.o was causing a parse error.

collectd module now skips this configuration for collectd for collectd == 5.9.0
Previously it was '!=' 5.9.

* Upsteam bug for collectd: https://github.com/voxpupuli/puppet-collectd/issues/901
* Previous incorrect commit: https://github.com/voxpupuli/puppet-collectd/commit/8edc01430d03eeabe812a7bab065f6b25200fda3

### DIFF
--- a/spec/acceptance/class_plugin_load_spec.rb
+++ b/spec/acceptance/class_plugin_load_spec.rb
@@ -1,0 +1,127 @@
+require 'spec_helper_acceptance'
+
+describe 'collectd::plugin::load class' do
+  context 'basic parameters' do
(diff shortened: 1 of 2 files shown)


bastelfreak:
thanks for the work!

traylenator:
@bastelfreak thanks - one more bug that I am aware of.

PR #929, opened 2020-03-31
Install disk package if required prior CentOS 8

NikolayTsvetkov:

This PR is a bug fix allowing to install the disk package in versions prior CentOS 8 in case `$manage_package` is configured by the user.


bastelfreak:
thanks for the fix!

PR #930, opened 2020-03-31
[collectd] rename log_parser to logparser

prabiegx:
Renaming log_parser to logparser to use correct name of collectd logparser plugin:
https://github.com/collectd/collectd/blob/d317751670410c94f17563c33c00cda9c1efba2e/src/logparser.c#L34


prabiegx:
Tests passed after rerun. @bastelfreak Can this be merged?

PR #933, opened 2020-04-06
Add backwards compatibility in transition to Stdlib::Host

amoralej:
Patch [1] changed all `port` parameters to use `Stdlib::Port`. When
moving from String type to the new type, this patch is backwards
incompatible leading to type errors when applying manifest. This patch
is using Variant type to make this change backwards compatible and ease
transition to new versions of puppet-collectd.

[1] https://github.com/voxpupuli/puppet-collectd/commit/d7b79c502bd10587d7498ddc1d19484eabfd7380

### DIFF
--- a/manifests/plugin/amqp1.pp
+++ b/manifests/plugin/amqp1.pp
@@ -104,7 +104,7 @@ class collectd::plugin::amqp1 (
   Boolean $manage_package            = $collectd::manage_package,
   String $transport                  = 'metrics',
   Stdlib::Host $host                 = 'localhost',
-  Stdlib::Port $port                 = 5672,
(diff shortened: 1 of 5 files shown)


bastelfreak:
Hi @amoralej,

We already made a backwards incompatible release for this so it's unlikely that we will fix this. It should not be a problem to update the local code before switching to our 12.0 release?

amoralej:
> Hi @amoralej,
> 
> We already made a backwards incompatible release for this so it's unlikely that we will fix this. It should not be a problem to update the local code before switching to our 12.0 release?

I'll close this PR, i already workarounded the issue locally and was able to move on.

PR #935, opened 2020-04-15
Collectd plugin fix

MichalRebisz:
This is a fix for an issue encountered when trying to use puppet on tripleo deployment with mcelog and snmp_agent plugin. 

The deployment would fail pointing at snmp_agent.pp and mcelog.pp. Changed the puppet files and templates to be more in line with the other plugins.


manifests/plugin/snmp_agent.pp
```diff
@@ -44,10 +44,7 @@
 
   collectd::plugin { 'snmp_agent':
     ensure  => $ensure,
-    content => epp('collectd/plugin/snmp_agent.conf.epp', {
-      'data'  => $data,
```
bastelfreak, line 48: I don't quite understand how this makes a difference. The variable `$data` and also `$table` are available in this context. Can you please provide an acceptance test for this?

MichalRebisz: Sure ill write some additional tests. As for how this makes a difference im not entirely sure I understand it myself to be honest. As far as I can make sense of it taking into account the error im getting. It looks like when it tries to get the $data out of this context it ends up being null for some reason and although data and table are optional in the template it break. So I tried to work around this by removing this part and getting the variables by a full path in the template. This seems to work correctly both on tripleo and the existing tests on puppet.

MichalRebisz:
After updating to puppet-collectd 12.0 on tripleo this seems no longer to be an issue. I will close this PR since it looks like the issue might have been caused by something else.

PR #940, opened 2020-05-13
Fix unixsock default path

smortex:

The default path for unixsock has been /var/run/collectd-unixsock since
the first version of the plugin, 14 years ago:
https://github.com/collectd/collectd/blob/4bf284162200a304f0ec8d1f5c2dd93339ddc64c/src/unixsock.c#L48

This path is not overridden by the Debian nor CentOS packages as
installed by the module.

Use this path by default so that collectdctl without argument can
connect to the socket, unbreaking the bolt tasks.

n/a


bastelfreak:
thanks for the fix!

PR #941, opened 2020-05-13
Add ubuntu 20 support

bastelfreak:

### DIFF
--- a/manifests/plugin/rabbitmq.pp
+++ b/manifests/plugin/rabbitmq.pp
@@ -62,7 +62,8 @@ class collectd::plugin::rabbitmq (
   include collectd
 
   if $facts['os']['family'] == 'Debian' and versioncmp($facts['os']['release']['major'], '11') >= 0 or
-    $facts['os']['family'] == 'RedHat' and versioncmp($facts['os']['release']['major'], '8') >= 0 {
+  $facts['os']['family'] == 'RedHat' and versioncmp($facts['os']['release']['major'], '8') >= 0 or
+  $facts['os']['name'] == 'Ubuntu' and versioncmp($facts['os']['release']['major'], '20.04') >= 0 {
     fail('https://pypi.org/project/collectd-rabbitmq/ does not support Python 3')
   }
 
--- a/spec/spec_helper_methods.rb
+++ b/spec/spec_helper_methods.rb
@@ -15,32 +15,6 @@ def os_specific_options(facts)
   end
 end
 
-def all_supported_os_hash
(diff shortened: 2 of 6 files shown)


mmoll:
No idea if the remaining problems are because of https://github.com/collectd/collectd/issues/3370 or something else :man_shrugging:.

spec/spec_helper_methods.rb
```diff
@@ -20,15 +20,15 @@ def all_supported_os_hash
     supported_os: [
```
ekohl, line 20: This whole method appears to be unused. Can't we drop it?

.sync.yml
```diff
@@ -1,3 +0,0 @@
----
```
ekohl, line 1: Is this really the right thing to do?

bastelfreak: a previous version of this PR did a cleanup of the old travis options. I updated the PR.

manifests/plugin/rabbitmq.pp
```diff
@@ -61,7 +61,7 @@
 ) {
   include collectd
 
-  if $facts['os']['family'] == 'RedHat' and versioncmp($facts['os']['release']['major'],'8') >= 0 {
+  if ($facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] == '8') or ($facts['os']['name'] == 'Ubuntu' and $facts['os']['release']['major'] == '20.04') {
```
ekohl, line 64: This means it no longer works for EL9 nor Fedora. While not technically supported, I think it's good to look to the future.

ekohl (commented):
Should we actually have a `Boolean $collectd::is_python3_os` to remove all the duplicated logic?

smortex:
I scratched my head with these tests too in #984 (they also now fail with old supported platforms).  A lot of these failing tests fail only due to an incompatible execution environment: each one is unique and nothing can be generalized :-/

The rspec-puppet tests in `spec/classes` / `spec/defines` already cover these aspects so I proposed to remove the problematic tests in #985.  I think that rebasing on top of it would help if you are okay with this.

manifests/plugin/rabbitmq.pp
bastelfreak: good catch, fixed it

bastelfreak:
> Should we actually have a `Boolean $collectd::is_python3_os` to remove all the duplicated logic?

I really like the idea but currently don't have the time to implement this :(

TuningYourCode (approved):

PR #942, opened 2020-05-13
Allow the getval task to work with customized hostnames

smortex:

The default hostname used by collectd is system dependent as retuned by
gethostname(3) (On FreeBSD it's the FQDN, while on Linux it's the
non-FQDN).  The python method socket.gethostname() always return a
non-FQDN, and the user has the ability to tune the hostname as he wish
in the collectd configuration (hostname parameter).

This change allow to use a customized hostname by using an optional
argument "hostname".  When not provided, the non-FQDN and the FQDN are
tried in turn to fetch the requested metric value.

n/a

### DIFF
--- a/tasks/getval.json
+++ b/tasks/getval.json
@@ -3,6 +3,10 @@
     "input_method": "stdin",
     "files": ["python_task_helper/files/task_helper.py"],
     "parameters": {
+      "hostname": {
+        "description": "Hostname of metric, as configured with Hostname in collectd",
+        "type": "Optional[String[1]]"
+      },
       "metric": {
         "description": "Name of metric, e.g. load/load-relative",
         "type": "String[1]"
--- a/tasks/getval.py
+++ b/tasks/getval.py
@@ -8,11 +8,24 @@ sys.path.append(os.path.join(os.path.dirname(__file__), '..', '..', 'python_task
 from task_helper import TaskHelper
 
 
-HOSTNAME = socket.gethostname()
 class GetVal(TaskHelper):
+    def get_metric(self, hostname, metric):
+        fullmetric = hostname+'/'+metric
+        return subprocess.check_output(['/usr/bin/collectdctl', 'getval',fullmetric]).rstrip().decode().split('\n')
+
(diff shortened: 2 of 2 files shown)


tasks/getval.json
```diff
@@ -3,6 +3,10 @@
     "input_method": "stdin",
     "files": ["python_task_helper/files/task_helper.py"],
     "parameters": {
+      "hostname": {
```
traylenator, line 6: Hi
There's example and parameters in README that could be extended with the new option.

traylenator:
Glad someone using this, I pretty much did it as an academic exercise

traylenator:
Rather than passing the `hostname` which will be different for every host can something be done with a fact name. Pass a factname and then that be used. 
Trouble is I can't remember much how this stuff works.

smortex:
Ah yes, didn't thing that passing the hostname does not scale beyond 1 node…  Maybe we can just use hostname, and fallback to fqdn (so no custom values).  We can also look for a setting in `/etc/collectd/collectd.conf` before falling back but some systems will use e.g. `/usr/local/etc/collectd/collectd.conf`…

Not ideal…  What do you think ?

traylenator:
Parsing collectd.conf seems  a bit yuck.

We use:
```
FQDNLookup true
```

Sending a fact name of choice as a parameter then the user can specify if they
want  networking.fqdn or networking.hostname

traylenator:
I never realised you do a getval without the hostname.

```
collectdctl getval ai-catalog-diff/gauge-return_code
```

works perfectly well, that's the solution.

smortex:
> I never realised you do a getval without the hostname.
Interesting, I do not see this behavior and was thinking about contributing it maybe…  A patch might be already available somewhere. On what system do you experience this?

traylenator:
rpm -q collectd
collectd-5.8.1-1.el7.cern8.x86_64

Don't think we have any patches in this area.

We do have:
```
FQDNLookup true
```

I wonder if your $(hostname) is not equaul to $(hostname -f) ?

PR #943, opened 2020-05-26
Remove redundant white space from python templates

traylenator:
Simplified the epp template contained:

```erb
    <%# a comment #%>
```

* In puppet 5 this renders as a new line. The prefix of spaces is eaten and a new line is created.
* In puppet 6 this renders as a string of spaces and new line.

The puppet 6 behaviour is better so I expect a bug was fixed sometime.

In reality there should be no spaces in here so change to

```erb
    <%#- a comment -%>
```

which renders as nothing and gives us zero catalogue change between 5 and 6 :-)

Extra commit also to remove a redundant blank line from `python-config.conf`


bastelfreak:
thanks for the fix!

PR #947, opened 2020-06-23
puppet-lint: Enable whitespace checker

bastelfreak:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,7 +11,8 @@ def location_for(place, fake_version = nil)
 end
 
 group :test do
-  gem 'voxpupuli-test', '>= 1.0.0',  :require => false
+  gem 'voxpupuli-test', '>= 1.4.0',  :require => false
+  gem 'puppet-lint-manifest_whitespace-check', :require => false
   gem 'coveralls',                   :require => false
   gem 'simplecov-console',           :require => false
 end
--- a/examples/plugins/ceph.pp
+++ b/examples/plugins/ceph.pp
@@ -1,7 +1,5 @@
 include collectd
 
 class { 'collectd::plugin::ceph':
-  osds          => [ 'osd.0', 'osd.1', 'osd.2'],
+  osds          => ['osd.0', 'osd.1', 'osd.2'],
 }
-
-
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -1,6 +1,5 @@
 #
 class collectd::params {
-
   $autoloadplugin            = false
   $fqdnlookup                = true
   $collectd_hostname         = $facts['networking']['hostname']
@@ -16,7 +15,7 @@ class collectd::params {
   $read_threads              = 5
   $write_threads             = 5
   $timeout                   = 2
-  $typesdb                   = [ '/usr/share/collectd/types.db' ]
+  $typesdb                   = ['/usr/share/collectd/types.db']
   $write_queue_limit_high    = undef
   $write_queue_limit_low     = undef
   $package_ensure            = 'present'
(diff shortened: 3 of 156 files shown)


bastelfreak:
Hi people,
please comment on the diff. I would like to add https://github.com/kuleuven/puppet-lint-manifest_whitespace-check to our testmatrix. The whole diff is caused by the autofix. I'm currently tracking down one bug: https://github.com/kuleuven/puppet-lint-manifest_whitespace-check/issues/3

bastelfreak:
upstream bug: https://github.com/kuleuven/puppet-lint-manifest_whitespace-check/issues/4

examples/plugins/ceph.pp
```diff
@@ -1,7 +1,5 @@
 include collectd
 
 class { 'collectd::plugin::ceph':
-  osds          => [ 'osd.0', 'osd.1', 'osd.2'],
+  osds          => ['osd.0', 'osd.1', 'osd.2'],
```
alexjfisher, line 4: I thought the linter was also going to pick up the crazy about of extra whitespace here?

manifests/plugin/apache.pp
```diff
@@ -2,11 +2,10 @@
 class collectd::plugin::apache (
   Enum['present', 'absent'] $ensure        = 'present',
   Boolean $manage_package                  = $collectd::manage_package,
-  Hash $instances                          = { 'localhost' => { 'url' => 'http://localhost/mod_status?auto' } },
+  Hash $instances                          = { 'localhost' => { 'url' => 'http://localhost/mod_status?auto' }},
```
alexjfisher, line 5: Why is this space removed, but not the others?

manifests/params.pp
```diff
@@ -33,12 +32,12 @@
 
   case $facts['kernel'] {
     'OpenBSD': { $has_wordexp = false }
-    default:   { $has_wordexp = true }
+    default: { $has_wordexp = true }
```
alexjfisher, line 35: It made sense to align the cases before though?

examples/plugins/ceph.pp
jovandeginste: I thought there was a different linter for this. I may be mistaken... Maybe it only verifies whether they are vertically aligned (which is always true when there is only one element)

manifests/params.pp
jovandeginste: Agreed

manifests/plugin/apache.pp
jovandeginste: As a side note, I had a similar case where I had a default value for a Hash (but more data and thus multiline), which crashed [class_parameter-check](https://rubygems.org/gems/puppet-lint-class_parameter-check)

examples/plugins/ceph.pp
bastelfreak: that should be fixed with this plugin: https://github.com/voxpupuli/puppet-collectd/pull/948

Gemfile
```diff
@@ -11,7 +11,8 @@ def location_for(place, fake_version = nil)
 end
 
 group :test do
-  gem 'voxpupuli-test', '>= 1.0.0',  :require => false
+  gem 'voxpupuli-test', '>= 1.4.0',  :require => false
+  gem 'puppet-lint-manifest_whitespace-check', :require => false
```
ekohl, line 15: I'd prefer to make this a part of vp-test 1.5.0 if it works well (which it appears to do)

bastelfreak: yes, me too. I just added it for testing / enable people to reproduce this.

ekohl:
This was included in modulesync 3.0.0

PR #948, opened 2020-06-23
puppet-lint: fix strict_indent

bastelfreak:
this check comes from a new plugin:
https://github.com/relud/puppet-lint-strict_indent-check

if this works fine I would like to add it as new dependency to
voxpupuli-test.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,7 @@ end
 
 group :test do
   gem 'voxpupuli-test', '>= 1.0.0',  :require => false
+  gem 'puppet-lint-strict_indent-check', :require => false
   gem 'coveralls',                   :require => false
   gem 'simplecov-console',           :require => false
 end
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -12,8 +12,7 @@ class collectd::install {
     }
   }
 
-  if $collectd::utils and  ( $facts['os']['family'] == 'Debian' or
-    ( $facts['os']['family'] == 'RedHat' and versioncmp($facts['os']['release']['major'],'8') >= 0 )) {
+  if $collectd::utils and  ( $facts['os']['family'] == 'Debian' or ( $facts['os']['family'] == 'RedHat' and versioncmp($facts['os']['release']['major'],'8') >= 0 )) {
     package{'collectd-utils':
       ensure => $collectd::package_ensure,
     }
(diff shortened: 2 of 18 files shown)


Gemfile
```diff
@@ -12,6 +12,7 @@ end
 
 group :test do
   gem 'voxpupuli-test', '>= 1.0.0',  :require => false
+  gem 'puppet-lint-strict_indent-check', :require => false
```
bastelfreak, line 15: I didn't adjust the indent here because it will be purged by the next modulesync anyways.

manifests/install.pp
```diff
@@ -13,7 +13,7 @@
   }
 
   if $collectd::utils and  ( $facts['os']['family'] == 'Debian' or
-    ( $facts['os']['family'] == 'RedHat' and versioncmp($facts['os']['release']['major'],'8') >= 0 )) {
+  ( $facts['os']['family'] == 'RedHat' and versioncmp($facts['os']['release']['major'],'8') >= 0 )) {
```
ekohl, line 16: This one looks odd to me

manifests/plugin/perl/plugin.pp
```diff
@@ -71,7 +71,7 @@
       }
     }
     default: {
-      fail("Unsupported provider: ${provider}. Use 'package', 'cpan',
+    fail("Unsupported provider: ${provider}. Use 'package', 'cpan',
```
ekohl, line 74: This also looks odd

bastelfreak:
@ekohl fixed all

manifests/plugin/disk.pp
```diff
@@ -17,9 +17,9 @@
       $_manage_package = $manage_package
     } else {
       if versioncmp($collectd::collectd_version_real, '5.5') >= 0
-        and versioncmp($facts['os']['release']['major'],'8') >= 0 {
+      and versioncmp($facts['os']['release']['major'],'8') >= 0 {
```
ekohl, line 20: This also looks incorrect. I think that's a bug in the actual lint plugin on multi line if conditions. Perhaps it needs braces?

bastelfreak: yes it need braces. I will add them and raise an issue upstream

bastelfreak: not sure if this an actual bug in the Puppet DSL. the code still compiles correctly?

ekohl: AFAIK Puppet DSL doesn't care about whitespace at all.

ekohl:
This was included in modulesync 3.0.0.

PR #949, opened 2020-06-26
Implement puppet-lint-class_parameter-check

bastelfreak:

### DIFF
--- a/manifests/plugin/aggregation.pp
+++ b/manifests/plugin/aggregation.pp
@@ -1,8 +1,8 @@
 # collectd::plugin::aggregation
 class collectd::plugin::aggregation (
-  Enum['present', 'absent'] $ensure = 'present',
-  Optional[Integer[1]] $interval    = undef,
   Hash $aggregators                 = {},
+  Enum['present', 'absent'] $ensure = 'present',
+  Optional[Integer[1]] $interval    = undef
 ) {
 
   include collectd
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,7 @@ end
 
 group :test do
   gem 'voxpupuli-test', '>= 1.0.0',  :require => false
+  gem 'puppet-lint-class_parameter-check', :require => false, :git => 'https://github.com/bastelfreak/puppet-lint-class_parameter-check', :branch => 'fix'
   gem 'coveralls',                   :require => false
   gem 'simplecov-console',           :require => false
 end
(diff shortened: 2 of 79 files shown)


bastelfreak:
This is a test for a plugin. To get this working we require https://github.com/ryreitsma/puppet-lint-class_parameter-check/pull/25 to be fixed. @voxpupuli/collaborators please let me know if you like the idea that parameters are sorted alphabetically

ekohl:
I generally don't like this because I'm worried about usefulness. For example, I like to keep `manage_db` together with `db_name`. Maybe it should be `db_manage` but renaming a parameter just so you can sort doesn't feel like a huge improvement.

manifests/plugin/aggregation.pp
```diff
@@ -1,8 +1,8 @@
 # collectd::plugin::aggregation
 class collectd::plugin::aggregation (
-  Enum['present', 'absent'] $ensure = 'present',
-  Optional[Integer[1]] $interval    = undef,
   Hash $aggregators                 = {},
+  Enum['present', 'absent'] $ensure = 'present',
```
ekohl, line 4: I believe there's another lint check that prefers ensure to be the first param. Perhaps that's only for instantiation though.

bastelfreak: correct, that's only for resources (and as far as I know thats active in our setup)

bastelfreak:
@ekohl I agree, personally I'm not 100% sure if I want this plugin or not, but I thought we gave it a try and let our collaborators decide.

traylenator:
Parameters are often sorted by usefulness as well I'd say or relevance.
```
users_allow
groups_allow
```
you would not want to split.

ekohl:
So what's the conclusion about this?

ghoneycutt:
@bastelfreak I like this for modules with only a few parameters. The functionality of just checking that parameters without defaults go before parameters with defaults would be very useful without the rest of the params being in alphabetic ordering. 

Take for instance a class that has a template with a lot of parameters in it. It is more useful to separate the class params by function so you might have params in alpha order that all do with the module like file name, service name, etc though the params that are used in templateA are all together and the params for templateB are all together.

ghoneycutt:
@bastelfreak Functionality aside, I'm not a fan of commits that re-order things or simply change their formatting, much like an aggressive rubocop auto fix. The reason is that it forces people to do some complicated rebases that could just turn them away from the project. If you are going to do a large amount of reformatting code, it is better to do when there are no big PR's in flight and to coincide with a major version bump. That will minimize the impact to developers.

bastelfreak:
since not a majority voted for this plugin I'm going to close the PR.

PR #950, opened 2020-07-23
Add options to limit the send queue length

mrunge:
Collectd has the option to limit the send queue lenght
for amqp1. This patch adds support for that.

For reference, that was added in
https://github.com/collectd/collectd/pull/3432/files

### DIFF
--- a/examples/plugins/filecount.pp
+++ b/examples/plugins/filecount.pp
@@ -7,11 +7,11 @@ class { 'collectd::plugin::filecount':
   },
 }
 
-collectd::plugin::filecount::directory {'foodir':
+collectd::plugin::filecount::directory { 'foodir':
   path => '/path/to/dir',
 }
 
-collectd::plugin::filecount::directory {'aborted-uploads':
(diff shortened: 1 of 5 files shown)


mrunge:
Does anyone have an idea why CI fails on this? I mean, I see a lot of "errors", but they are not related to this change?

mrunge:
Is there a way to test this patch on top of another PR, like on https://github.com/voxpupuli/puppet-collectd/pull/951 ?

examples/plugins/filecount.pp
```diff
@@ -7,11 +7,11 @@
   },
 }
 
-collectd::plugin::filecount::directory { 'foodir':
+collectd::plugin::filecount::directory {'foodir':
```
bastelfreak, line 10: I think out linter plugin requires the whitespace here?

mrunge: yes! I thought I had that commit removed. This has nothing to do with the original change and was just a try to make CI pass.

PR #955, opened 2020-09-11
Add PluginInstanceFormat parameter to virt plugin

leifmadsen:
Add the PluginInstanceFormat parameter to the virt plugin to allow for specifying the
plugin instance format, resulting in a new label being available to the virt metrics. By
adding this you can provide the UUID of the virtual machine, along with HostnameFormat
to have information about VMs that exist per physical host.

### DIFF
--- a/manifests/plugin/virt.pp
+++ b/manifests/plugin/virt.pp
@@ -1,17 +1,18 @@
 # https://collectd.org/wiki/index.php/Plugin:virt
 class collectd::plugin::virt (
   String $connection,
-  $ensure                                      = 'present',
-  $manage_package                              = undef,
-  Optional[Pattern[/^\d+$/]] $refresh_interval = undef,
-  Optional[String] $domain                     = undef,
-  Optional[String] $block_device               = undef,
-  Optional[String] $interface_device           = undef,
-  Optional[Boolean] $ignore_selected           = undef,
-  Optional[String] $hostname_format            = undef,
-  Optional[String] $interface_format           = undef,
-  Optional[String] $extra_stats                = undef,
-  $interval                                    = undef,
+  $ensure                                                                    = 'present',
+  $manage_package                                                            = undef,
+  Optional[Pattern[/^\d+$/]] $refresh_interval                               = undef,
+  Optional[String] $domain                                                   = undef,
+  Optional[String] $block_device                                             = undef,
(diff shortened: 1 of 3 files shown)


igalic (changes requested):
this needs a spec test

leifmadsen:
@igalic test added

templates/plugin/virt.conf.erb
```diff
@@ -19,6 +19,11 @@
 <% if @ignore_selected != nil -%>
   IgnoreSelected <%= @ignore_selected %>
 <% end -%>
+<%- if scope.lookupvar('collectd::collectd_version_real') and (scope.function_versioncmp([scope.lookupvar('collectd::collectd_version_real'), '5.5']) >= 0) -%>
```
igalic, line 22: i can't tell why yet, but i don't like this line one bit

igalic (commented):
🤔

templates/plugin/virt.conf.erb
leifmadsen: That's fair :) I only added it because I was trying to get the tests to make sense. The `PluginInstanceFormat` was only added as of 5.5.0 fully, and the template would include it when generating the resulting configuration, which was causing the tests to fail.

I haven't used puppet in several years, so I'm honestly not sure the appropriate approach here.

leifmadsen: Any suggestions on what I should do here if anything?

igalic: no
not before we convert this epp

manifests/plugin/virt.pp
```diff
@@ -8,6 +8,7 @@
   Optional[String] $block_device               = undef,
   Optional[String] $interface_device           = undef,
   Optional[Boolean] $ignore_selected           = undef,
+  Optional[String] $plugin_instance_format     = undef,
```
bastelfreak, line 11: I guess only certain values are allowed as format? If so, please provide them as an enum

templates/plugin/virt.conf.erb
bastelfreak: yes it's ugly, but erb is pretty limited :sadface:

bastelfreak (approved):

PR #957, opened 2020-09-24
Fix `ignoreselected` typo in protocols.conf.erb template

VtG242:

### DIFF
--- a/templates/plugin/protocols.conf.erb
+++ b/templates/plugin/protocols.conf.erb
@@ -9,8 +9,8 @@ if @values
   end
 end
 -%>
-<% if @ignore_selected != nil -%>
-  IgnoreSelected <%= @ignore_selected %>
+<% if @ignoreselected != nil -%>
+  IgnoreSelected <%= @ignoreselected %>
 <% end -%>
 </Plugin>
 <% end -%>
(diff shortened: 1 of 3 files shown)


kenyon (approved):

igalic:
why are unit Tests failing on puppet 5 & 6?

templates/plugin/protocols.conf.erb
```diff
@@ -9,8 +9,8 @@ if @values
   end
 end
 -%>
-<% if @ignore_selected != nil -%>
-  IgnoreSelected <%= @ignore_selected %>
+<% if @ignoreselected != nil -%>
```
alexjfisher, line 12: Currently, `ignoreselected` is a Boolean.  That means it can never be `nil` (undef in puppet).

alexjfisher: Looks like it was always a boolean and never default to `undef`.

bastelfreak (approved):

alexjfisher:
@VtG242 Thanks for the bugfix!  I've made the existing test pass by changing the default to `undef` and added extra tests for the cases it's explicitly set to `true` or `false`.

Issue #958, opened 2020-10-12
ovs_stats port number mismatch


mrunge:
The port number for ovs_stats port is a string, see 
https://github.com/collectd/collectd/blob/main/src/collectd.conf.in#L1339

in in puppet-collectd, it is a port
https://github.com/voxpupuli/puppet-collectd/blob/master/manifests/plugin/ovs_stats.pp#L42

Ticks are missing around https://github.com/voxpupuli/puppet-collectd/blob/master/templates/plugin/ovs_stats.conf.erb#L3
(port).

PR #959, opened 2020-10-12
Fix type mismatch for ovs_stats port

mrunge:

There is a typo in the template to create the ovs_stats config. For collectd, the given port number is a string, where
puppet-collectd treats this as a port, thus resulting as a number in the collectd config file.
Since collectd upstream uses 6640 by default, this option is rarely required.

Fixes: #958


bastelfreak:
@mrunge can you take a look at the failing unit tests and maybe add an acceptance test for this change? We've a bunch of examples at https://github.com/voxpupuli/puppet-collectd/tree/master/spec/acceptance

mrunge:
Thank you for your quick look.

The failing tests lead me to the acceptance tests. If the tests are passing now, this should be good to go :)

mrunge:
Okay, tests pass now and the tests also check for existence of quotation marks.

bastelfreak:
can you please add an acceptance test for this as well? :)

mrunge:
Tests fail with:
 E: Unable to locate package collectd-ovs-stats on ubuntu

mrunge:
The package is not available on ubuntu and debian (which uses the same repo); it also fails to load on centos either.
There is not much value in adding a test, when the infra can not check it properly.

mrunge:
@bastelfreak any opinion to move forward here?

PR #960, opened 2020-10-22
Fix dbi query.conf.erg template

VtG242:
Without this change collectd refuse to load configuration for dbi plugin with error (at least for me - collectd-5.8.1):
The Query block needs exactly one string argument.
dbi plugin: No <Query> blocks have been found. Without them, this plugin can't do anything useful, so we will return an error.

### DIFF
--- a/templates/plugin/dbi/query.conf.erb
+++ b/templates/plugin/dbi/query.conf.erb
@@ -1,4 +1,4 @@
-  <Query <%= @name %>>
+  <Query "<%= @name %>">
     Statement "<%= @statement %>"
 <% if @minversion -%>
     MinVersion <%= @minversion %>


bastelfreak:
@VtG242 thanks for the patch! Any chance to write a simple acceptance test that valides the new config?

VtG242:
To be honest I don't have much experience with this. 
If there is some documentation how to do it I could try ;-)

bastelfreak:
Hey,
can you take a look at the different test files in https://github.com/voxpupuli/puppet-collectd/tree/master/spec/acceptance ?
for each of them we spin up multiple docker containers with centos/debian/ubuntu, apply the puppet code from the test file and afterwards the checks. there are already a few files that test collectd plugins. We had multiple quoting issues in the past and it's always bad when we change it back and forth. So if we have a very simple test that just checks if collectd successfully loaded the config file, that would already be sufficient. you can copy one of the existing files, adjust it for the dbi plugin and commit it. Our CI setup will automatically test if. 

If you've any further questions, don't hesitate to ping us in the IRC channel #voxpupuli on freenode or the slack room #voxpupuli on https://slack.puppet.com/

VtG242:
Ok I will take a look at it once I will have a little spare time.

VtG242:
I could write test, but I need support within docker images. For real test that collectd is running with dbi plugin enabled I need any db engine there (e.g. sqlite3 for simplicity) with some simple database and table with data e.g.:
` sqlite3 /opt/dbi-test.db 'CREATE TABLE TEST(ID INT);INSERT INTO TEST (ID) VALUES ('1');'`
Is it possible?

bastelfreak:
it's a bit annoying that collect not simply enables the plugin :( We could setup databases during the tests, but that's probably not worth the hassle for this tiny fix.

PR #965, opened 2020-12-17
Separation of values for a python plugin array module key

### LINKED ISSUE #964: Python plugin template does not create valid module configuration

- Puppet: 6.18
- Ruby: 2.6.2-118
- Distribution: Centos 7.9
- Module version: 12.2.1-rc0

Create python plugin module configuration eg.
```
collectd::plugin::python::module {
  'haproxy':
     config  => [
       { 
         'Verbose'      => $verbose_logging,
         'ProxyMonitor' => $proxy_monitor,
         'ProxyIgnore'  => $proxy_ignore,
         'Socket'       => $socket,
       }
     ],
```
Using the https://github.com/mleinart/collectd-haproxy as a module for python plugin. 

## What are you seeing

The generated configuration (more values assigned to the key on a single line separated by space) eg.
```
ProxyMonitor "frontend" "backend" "server"
```
is not compatible with collectd [register_config](https://collectd.org/documentation/manpages/collectd-python.5.shtml#register_config) function using the [Config](https://collectd.org/documentation/manpages/collectd-python.5.shtml#config) object where it is written.

> The sequence of children keeps one entry for each configuration option.
> -- <cite>https://collectd.org/documentation/manpages/collectd-python.5.shtml#config</cite>

So the array for values of module `key` cannot be formed like that as it is not propagated to the module by collectd itself.

The single value per module configuration key is expected, so the result is supposed to be like.
```
ProxyMonitor "frontend"
ProxyMonitor "backend"
ProxyMonitor "server"
```

## Any additional information you'd like to impart

The affected part of the code is https://github.com/voxpupuli/puppet-collectd/blob/master/templates/plugin/python/module.conf_config.epp#L11-L21

Slabimic:
Collectd expects to have a single value per each key when it is not another Config object instance included within the module configuration. So the result from an array of params is supposed to be like.

Key "Value1"
Key "Value 2"
Key 22
Key false

Change of module configuration parsing array values to per single value fixes the issue.

Fixes #964

### DIFF
--- a/spec/classes/collectd_plugin_python_spec.rb
+++ b/spec/classes/collectd_plugin_python_spec.rb
@@ -150,7 +150,9 @@ describe 'collectd::plugin::python', type: :class do
               target: "#{options[:plugin_conf_dir]}/python-config.conf"
             )
             is_expected.to contain_concat__fragment('collectd_plugin_python_conf_alpha_config').with(content: %r{Beta 4.2})
-            is_expected.to contain_concat__fragment('collectd_plugin_python_conf_alpha_config').with(content: %r{Delta "a" 4 5})
+            is_expected.to contain_concat__fragment('collectd_plugin_python_conf_alpha_config').with(content: %r{Delta "a"})
+            is_expected.to contain_concat__fragment('collectd_plugin_python_conf_alpha_config').with(content: %r{Delta 4})
+            is_expected.to contain_concat__fragment('collectd_plugin_python_conf_alpha_config').with(content: %r{Delta 5})
             is_expected.to contain_concat__fragment('collectd_plugin_python_conf_alpha_config').with(content: %r{Gamma "b"})
           end
         end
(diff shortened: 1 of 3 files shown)


kenyon (approved):

Slabimic:
Hello guys @kenyon, @bastelfreak can we proceed with the revert? There was a misunderstanding in the single value for a key meaning because in the linked documentation in https://github.com/voxpupuli/puppet-collectd/issues/964 the 'Config' object passed can contain tuples of values (so it can contain more values per key in a single line). 
Due to this fact, the check (plugin) has no support for the current situation and should be modified instead of collectd configuration template.

kenyon:
@Slabimic did I miss an issue that was created or some discussion about reverting this?

Slabimic:
> @Slabimic did I miss an issue that was created or some discussion about reverting this?

@kenyon no issue was created, but I've got a better understanding of collectd-python documentation with representing tuple of values as one entry for the configuration key in Config object. So the previous template worked fine due to documentation and probably older imported plugins should be modified instead of the configuration template itself.
Maybe there was a discussion missing about that in the https://github.com/voxpupuli/puppet-collectd/issues/964.

kenyon:
@Slabimic Ok, so we just need to revert this pull request?

Slabimic:
> @Slabimic Ok, so we just need to revert this pull request?

@kenyon I think it would not cause any harm to make new features work again with the revert and I am sorry for enforcing the old configuration syntax method.

kenyon:
@Slabimic see #967.

PR #970, opened 2021-05-29
MySQL database plugin: allow no password; default Host, User, and Password to undef

### LINKED ISSUE #724: plugin::mysql::database - remove default 'UNSET' values
Use better types to make it clearer that these parameter are required

ph448:
This started off as a change to allow for not specifying a password, as my setup uses socket authentication, then I noticed #724 and thought to include the changes to fix that as well.
Setting the default host to 'localhost' and not requiring a username or password is better in line with `mysqlclient` defaults, which the collectd mysql plugin ultimately uses to interact with the server. Also, the plugin supports specifying a socket, for which no username or password is required, so having them optional allows for configuring the plugin in correct ways that were not possible before the change.
Fixes #724


smortex (commented):
That looks reasonable.  I guess empty strings are not valid for these parameters?  If so, `String[1]` should be preferred.

ph448:
@smortex empty strings are probably not valid for any of the string parameters in `manifests/plugin/mysql/database.pp`, but none of them have a minimum length requirement specified, making it look like a convention rather than an omission. that said, I'm happy to make the change, shall I change all of them?

smortex:
This module has a lot of classes and a rather long history, it has some classes without data types at all, some with "relaxed" data types and some with "strict" data types.  Tackling all this at once is not something appealing and fun, and I think that improving bits here and there is a better way to go, progressively improving the module validation.

If MySQL allow empty password the module must allow this (no idea, I am not a user of MySQL), for the two other parameters I would go for a minimum length of 1.

Thanks!

ph448:
OK, I've added the length restriction to both `$username` and `$password` as it doesn't make sense to set either to an empty string. Seeing that `$port` was already using a puppetlabs-stdlib type I've chosen to use `Stdlib::Host` for `$hostname`.

ghoneycutt (approved):

smortex (approved):

PR #973, opened 2021-06-10
Adding Reverse Host parameter

markasammut:
In version 5.9.0 of collectd, a ReverseHost parameters was added to the write_graphite plugin.  If set to **true**, the (dot separated) parts of the host field of the value list will be rewritten in reverse order. The rewrite happens before special characters are replaced with the EscapeCharacter.

Feature is currently unavailable when using the module


xneo64:
Thanks for this feature implementation! Looking forward for it to be merged.

kenyon (approved):

PR #985, opened 2021-09-09
Update from voxpupuli modulesync_config

smortex:


smortex:
I can't reproduce the Debian / Ubuntu failure and they seem to be transient :-/

smortex:
These unreliable buggy tests are causing more problems when adding support for new OS (e.g. #941 ,
#984).

I am therefore proposing to get rid of them.

PR #992, opened 2021-10-26
Added dropduplicatefields for write_graphite

caramarc:


manifests/plugin/write_graphite/carbon.pp
```diff
@@ -1,19 +1,20 @@
...
-  Boolean $separateinstances = false,
-  Boolean $logsenderrors     = true,
-  Integer $reconnectinterval = 0,
-  Boolean $reversehost       = false,
+  $ensure                      = 'present',
+  Stdlib::Host $graphitehost   = 'localhost',
+  Stdlib::Port $graphiteport   = 2003,
+  Boolean $storerates          = true,
+  $graphiteprefix              = 'collectd.',
+  $graphitepostfix             = undef,
```
traylenator, line 8: Could you add something to the existing tests in spec/defines/collectd_plugin_write_graphite_spec.rb to exercise this new parameter.

traylenator (changes requested):

PR #997, opened 2022-05-05
Drop Arch Linux support

bastelfreak:
Arch doesn't have the collectd package in the official repos anymore.


bastelfreak:
not marking as breaking change because the package is already removed from the official repos. we just stop declaring official puppet support.

PR #998, opened 2022-06-23
Bump EPEL dependency to '< 5' for EL9 support

jovandeginste:
Signed-off-by: Jo Vandeginste <Jo.Vandeginste@kuleuven.be>

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -22,7 +22,7 @@
     },
     {
       "name": "puppet-epel",
-      "version_requirement": ">= 3.0.0 < 4.0.0"
+      "version_requirement": ">= 3.0.0 < 5.0.0"
     },
     {
       "name": "puppetlabs-python_task_helper",
@@ -34,21 +34,24 @@
       "operatingsystem": "RedHat",
       "operatingsystemrelease": [
         "7",
-        "8"
+        "8",
+        "9"
       ]
     },
     {
       "operatingsystem": "CentOS",
       "operatingsystemrelease": [
         "7",
-        "8"
+        "8",
+        "9"
       ]
     },
     {
       "operatingsystem": "OracleLinux",
       "operatingsystemrelease": [
         "7",
-        "8"
+        "8",
+        "9"
       ]
     },
     {
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -11,7 +11,7 @@ class collectd::install {
     }
   }
 
-  if $collectd::utils and  ( $facts['os']['family'] == 'Debian' or ( $facts['os']['family'] == 'RedHat' and versioncmp($facts['os']['release']['major'],'8') >= 0 )) {
+  if $collectd::utils and  ( $facts['os']['family'] == 'Debian' or ( $facts['os']['family'] == 'RedHat' and versioncmp($facts['os']['release']['major'],'8') == 0 )) {
     package { 'collectd-utils':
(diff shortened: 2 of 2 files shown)


metadata.json
```diff
@@ -32,51 +32,31 @@
   "operatingsystem_support": [
     {
       "operatingsystem": "RedHat",
-      "operatingsystemrelease": [
```
kajinamit, line 35: Can we keep the original multi-line format ? Main concern is that this is touching unrelated lines because of the format change.

kajinamit (changes requested):

kajinamit:
It seems collectd is not yet released in EPEL 9 and that's why the CentOS 9 tests are failing.

metadata.json
jovandeginste: Yes - my auto-formatter does this. I added a "plugin" that keeps arrays multiline now.

jovandeginste:
> It seems collectd is not yet released in EPEL 9 and that's why the CentOS 9 tests are failing.

So we wait until it lands, one way or another...

metadata.json
kajinamit: thanks. looks better now !

metadata.json
```diff
@@ -22,7 +22,7 @@
     },
     {
       "name": "puppet-epel",
-      "version_requirement": ">= 3.0.0 < 4.0.0"
```
bastelfreak, line 25: can you please move this commit to a new PR? allowing newer epel isn't related to implementing EL9 support.

jovandeginste: > can you please move this commit to a new PR? allowing newer epel isn't related to implementing EL9 support.

puppet-epel only supports EL9 sinds [v4.1.0](https://github.com/voxpupuli/puppet-epel/blob/master/CHANGELOG.md#v410-2022-03-26); I feel this PR is senseless without updating this...

jovandeginste: Also, it (bump epel) is the very title of this PR... I must be too confused by the heat and it being a Friday afternoon.

traylenator: If 4.1.0 is minumum for EPEL9 then the lower bound should be increased also accordingly.

As for if this it should be in this MR at all I think its okay.

traylenator:
collectd is now in EPEL 9 it seems so this is good to go once comments are addressed.

metadata.json
jovandeginste: > If 4.1.0 is minumum for EPEL9 then the lower bound should be increased also accordingly.

I understand the idea, but I'm hesitant here - you are not forced to use EL9, so maybe you want to keep using your current module versions. I know what a hell it is to try and update your modules when they all have some variation of requirements of a shared module (like epel).

Anyway, do you want me to increment the lower bound?

bastelfreak: this PR contains two changes. Adding EL 9 and allowing the latest EPEL module. Allowing  the latest EPEL module is a requirement for adding EL 9 support. But the change is important on it's own. the collectd module needs to allow the latest EPEL version in case it gets installed into an environment where other modules already require EPEL 4. because of that we should split this PR up into two. one allowing EPEL 4, the other adding code for EL 9.

jovandeginste:
Create two new PRs to split this one:

- https://github.com/voxpupuli/puppet-collectd/pull/1004
- https://github.com/voxpupuli/puppet-collectd/pull/1005

PR #1001, opened 2022-09-01
rubocop compliance for modulesync_config 5.3.0

traylenator:
rubocop corrections for 5.3.0

Includes the ignoring of RSpec/MultipleMemoizedHelpers which is very prevalent in this module.


.rubocop.yml
```diff
@@ -2,5 +2,6 @@
 # Managed by modulesync - DO NOT EDIT
 # https://voxpupuli.org/docs/updating-files-managed-with-modulesync/
 
+inherit_from: .rubocop_todo.yml
```
smortex, line 5: This will vanish on next modulesync update.

.rubocop_todo.yml
```diff
@@ -0,0 +1,4 @@
+---
+# Is very common in this module
+RSpec/MultipleMemoizedHelpers:
+  Enabled: false
```
smortex, line 4: I had to disable this from time to time, it does not work well with large code base. Maybe we can opt-out by default?

smortex (commented):
LGTM

.rubocop.yml
traylenator: Looking at https://github.com/voxpupuli/modulesync_config/blob/master/moduleroot/.rubocop.yml.erb#L5

```erb
<% if File.exist?(File.join(@metadata[:workdir], '.rubocop_todo.yml')) -%>
inherit_from: .rubocop_todo.yml
<% end -%>
```

I think not:

bastelfreak (approved):

.rubocop.yml
smortex: w00t! :heart_eyes:

smortex (approved):

PR #1007, opened 2022-12-02
Update socket path of ceph services in ceph plugin

yadneshk:
Ceph 16 onwards sockets for Ceph services are placed under '/var/run/ceph/\<fsid\>/'

This change introduces a new variable `ceph_fsid` to hold Ceph cluster's fsid which must be of type string.

Configuration file for ceph plugin will be generated with new path only if `ceph_fsid` is set.

n/a

Signed-off-by: Yadnesh Kulkarni <ykulkarn@redhat.com>


paramite:
+1

smortex (approved):
LGTM!

PR #1009, opened 2022-12-27
ValuesFrom in postgresql query.conf should allow multiple values

markasammut:
ValuesFrom in postgresql query.conf should allow an array of multiple values, not just a single value, as per documentation:
https://collectd.org/documentation/manpages/collectd.conf.5.shtml#query_blocks

Issue is not documented yet

### DIFF
--- a/templates/plugin/postgresql/query.conf.erb
+++ b/templates/plugin/postgresql/query.conf.erb
@@ -20,7 +20,13 @@
 <% if result['instancesfrom'] -%>
         InstancesFrom "<%= result['instancesfrom'] %>"
 <% end -%>
+<%- if result['valuesfrom'].is_a?(Array) -%>
+<%- result['valuesfrom'].each do |v| -%>
+        ValuesFrom "<%= v %>"
+<%- end -%>
+<%- else -%>
         ValuesFrom "<%= result['valuesfrom'] %>"
-      </Result>
+<%- end -%>
+    </Result>
 <% end -%>
   </Query>


templates/plugin/postgresql/query.conf.erb
```diff
@@ -20,7 +20,13 @@
 <% if result['instancesfrom'] -%>
         InstancesFrom "<%= result['instancesfrom'] %>"
 <% end -%>
+<%- if result['valuesfrom'].is_a?(Array) -%>
+<%- result['valuesfrom'].each do |v| -%>
+        ValuesFrom "<%= v %>"
+<%- end -%>
+<%- else -%>
```
smortex, line 27: Maybe more readable to wrap the value in an array and flatten it?

```suggestion
<%- [result['valuesfrom']].flatten.each do |v| -%>
        ValuesFrom "<%= v %>"
```

templates/plugin/postgresql/query.conf.erb
```diff
@@ -20,7 +20,13 @@
...
         ValuesFrom "<%= result['valuesfrom'] %>"
-      </Result>
+<%- end -%>
+    </Result>
```
smortex, line 30: Indend seems to not match the opening tag anymore:

```suggestion
      </Result>
```

markasammut:
Suggestions committed - removed extra line repeating the ValuesFrom

bastelfreak:
@markasammut thanks for the PR. can you please add a unit test for it?

markasammut:
Never really mastered writing one, but will try.

markasammut:
@bastelfreak modified spec file to test accepting an arrray - 1 CentOS8 check is failing but I don't believe it's related to this change.

PR #1010, opened 2022-12-27
Ensure plugin version is same version as main collectd package

### LINKED ISSUE #430: Plugins should have a default of collectd::package_ensure, not their own
If I set collectd::package_ensure to latest, it should set the default for all of the plugins as well.

markasammut:
If you attempt to specify the collectd version in Hiera, and a newer version is available, issue might be caused because the package install of the plugins on new hosts would install the latest version for both the plugin and the main package (as a dependency). Puppet then starts failing attempting to downgrade the main package to your specified version.

With this PR it is ensure that mysql, snmp, postgresql and curl, are installed with the same version as the core package. Same thing was done for the disk plugin some time ago: https://github.com/voxpupuli/puppet-collectd/blob/master/manifests/plugin/disk.pp#L26L37

Partially fixes #430

### DIFF
--- a/manifests/plugin/curl.pp
+++ b/manifests/plugin/curl.pp
@@ -10,9 +10,15 @@ class collectd::plugin::curl (
   $_manage_package = pick($manage_package, $collectd::manage_package)
 
   if $facts['os']['family'] == 'RedHat' {
+    if $ensure == 'present' {
+      $ensure_real = $collectd::package_ensure
+    } elsif $ensure == 'absent' {
+      $ensure_real = 'absent'
+    }
+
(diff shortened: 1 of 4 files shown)


markasammut:
@smortex @traylenator any idea why this might be failing CentOS 8 tests? Does not seem to be a problem with the Puppet code. Same failure is being experienced on https://github.com/voxpupuli/puppet-collectd/pull/1009

smortex (approved):
LGTM. I re-run the failed CI jobs and they no pass. Looks like the CentOS 8 images where having trancient failures.

PR #1011, opened 2023-01-11
Add support for the Host option for the write_prometheus plugin

gerases:
The option was introduced in version 5.9 of collectd

### DIFF
--- a/manifests/plugin/write_prometheus.pp
+++ b/manifests/plugin/write_prometheus.pp
@@ -1,5 +1,6 @@
 class collectd::plugin::write_prometheus (
   Stdlib::Port $port = 9103,
+  Optional[Stdlib::Host] $host = undef,
   $ensure = 'present',
 ) {
   include collectd
--- a/manifests/plugin/write_prometheus.pp
+++ b/manifests/plugin/write_prometheus.pp
@@ -1,3 +1,11 @@
+# Class: collectd::plugin::write_prometheus
+#
+# @see https://collectd.org/wiki/index.php/Plugin:Write_Prometheus
+
+# Configures write_prometheus plugin.
+#
+# @param port Stdlib::Port Defines the port on which to accept scrape requests from Prometheus. Default: 9103
+# @param ip Optional[Stdlib::IP::Address] Defines the IP address to bind to. In not specified, the listener will bind to all IPs present.
 class collectd::plugin::write_prometheus (
   Stdlib::Port $port = 9103,
   Optional[Stdlib::IP::Address] $ip = undef,
--- a/manifests/plugin/write_prometheus.pp
+++ b/manifests/plugin/write_prometheus.pp
@@ -4,8 +4,8 @@
 
 # Configures write_prometheus plugin.
 #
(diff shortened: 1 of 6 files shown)


spec/classes/collectd_plugin_write_prometheus_spec.rb
```diff
@@ -12,14 +12,51 @@
...
+            facts.merge(collectd_version: '5.9')
+          end
+
+          is_expected.to contain_file('write_prometheus.load').with(
+            ensure: 'present',
+            path: "#{options[:plugin_conf_dir]}/10-write_prometheus.conf",
+            content: %r{Port "9103"}
+          ).with(
+            content: %r{Host "abc"}
+          )
```
smortex, line 42: This should be wrapped in a `it` block:

```ruby
it 'description' do
  is_expected.to ...
end
```

gerases: Yeah, I'm still trying to set up my environment to run the tests properly. Work in progress. I should be back :)

gerases: Let me know how it's looking now

smortex (approved):
Looks good. I thing the documentation should rather be added as comments so that it end-up in REFERENCE.md, but the whole module needs update in regard to this.

gerases:
> Looks good. I thing the documentation should rather be added as comments so that it end-up in REFERENCE.md, but the whole module needs update in regard to this.

Cool, thank you!

manifests/plugin/write_prometheus.pp
```diff
@@ -1,5 +1,6 @@
 class collectd::plugin::write_prometheus (
   Stdlib::Port $port = 9103,
+  Optional[Stdlib::IP::Address] $ip = undef,
```
bastelfreak, line 3: please document the new parameter in the class with puppet-string style:
```
# @param ip this does blabla
```

gerases: Sure will do that Tue or so

manifests/plugin/write_prometheus.pp
```diff
@@ -1,5 +1,14 @@
+# Class: collectd::plugin::write_prometheus
+#
+# @see https://collectd.org/wiki/index.php/Plugin:Write_Prometheus
+
+# Configures write_prometheus plugin.
+#
+# @param port Stdlib::Port Defines the port on which to accept scrape requests from Prometheus. Default: 9103
+# @param ip Optional[Stdlib::IP::Address] Defines the IP address to bind to. In not specified, the listener will bind to all IPs present.
```
bastelfreak, line 8:
```suggestion
# @param port Defines the port on which to accept scrape requests from Prometheus. Default: 9103
# @param ip Defines the IP address to bind to. In not specified, the listener will bind to all IPs present.
```

gerases: I was using the SNMP plugin documentation as the example. I'm ok with this change.

PR #1030, opened 2023-10-23
Accept + in collectd_version Fact

quartsize:
The version string resulting from a `make dist` of collectd may contain a +, e.g. 5.12.0.167.g7c5ce9f+, from being run on a dirty working tree.


kenyon (approved):
Would be good to also add a test where the version has a `+` to ensure that this doesn't break in the future.

quartsize:
> Would be good to also add a test where the version has a `+` to ensure that this doesn't break in the future.

Added.

PR #1075, opened 2026-09-13
replace deprecated calls with Facter::Core::Execution

corporate-gadfly:


lib/facter/collectd_version.rb
```diff
@@ -10,8 +10,8 @@
 #
 Facter.add(:collectd_version) do
   setcode do
-    if Facter::Util::Resolution.which('collectd')
-      collectd_help = Facter::Util::Resolution.exec('collectd -h')
```
bastelfreak, line 14: in the future we should switch to `confine` here.

bastelfreak (approved):

lib/facter/collectd_version.rb
corporate-gadfly: Yes. I noticed that in splunk as well.

corporate-gadfly: E.g.:
```
    cmd = if File.exist?('C:/Program Files/Splunk/bin/splunk.exe')
            '"C:/Program Files/Splunk/bin/splunk.exe" --version'
          elsif File.exist?('/opt/splunk/bin/splunk')
            '/opt/splunk/bin/splunk --version --accept-license --answer-yes --no-prompt'
          end
    if cmd
      output = Facter::Util::Resolution.exec(cmd)
      if output =~ %r{^Splunk ([0-9.]+) \(} # rubocop:disable Style/IfUnlessModifier
        value = Regexp.last_match(1)
      end
    end
```

corporate-gadfly: * https://github.com/voxpupuli/puppet-collectd/pull/1076
* https://github.com/voxpupuli/puppet-authselect/pull/25

I gave up on the splunk one.

